PR #43, opened 2019-07-11
Add free functions to calculate the instersection as well as the union of two Polygons

AKreutz:
Tests are included and passing.

@klemense1


AKreutz:
I am no longer using these functions.

Issue #51, opened 2019-07-23
A lane and its successor lane are sometimes not connected


AKreutz:
When working with the file `4way_intersection.xodr` I noticed that in some cases there are gaps between a lane and the next succeeding lane. This can be seen here:
![successor_lanes](https://user-images.githubusercontent.com/18612662/61721406-b36d4e80-ad68-11e9-9034-c3aba7cadefa.png)
According to the roadgraph, lines of the same color are in a predecessor/successor relationship. For the (global) LaneIds 35->3, 37->10, 46->29 and 47->50 there is a gap between the corresponding lines.
A test case for this error is implemented in #52
@klemense1

klemense1:
commented in https://github.com/bark-simulator/bark/issues/73
closing

PR #52, opened 2019-07-23
A test case to detect an error in 4way_intersection.xodr

AKreutz:
Test case for issue #51


AKreutz:
Issue #51 has been closed.

PR #120, opened 2019-09-20
Map_interface: confirm TODO in rtree_lane building

xmyqsh:


juloberno:
the center line of the lane should be used for the rtree

Issue #148, opened 2019-10-21
Coding Guidelines are undefined


klemense1:
- define them (Google Style Guide?)
- implement them
- use cpplint and pythonlint to enforce them? --> for that, it has to be integrated into the bazel build process, otherwise nobody will use it

patrickhart:
Also additionally, I think that we should agree on a two-space indent everywhere (also in Python).

klemense1:
implemented cpplint format file for autoformatting according to google style guide in  #160 
file can be created as follows: https://leimao.github.io/blog/Clang-Format-Quick-Tutorial/

PR #153, opened 2019-10-21
added pylint and autopep8

klemense1:


klemense1:
partly solves #148 as we can now use formatting in vs code out of the box.

PR #179, opened 2019-11-28
Fix behavior models

cirrostratus1:
Fixes problems in:

`BehaviorLongitudinalAcceleration`:

* If `delta_time` is evenly divisible by the `integration_time`, the last point in the trajectory has its time advanced, but for the rest of the state vector `integration_time = 0` was used.

`BehaviorLongitudinalAcceleration`:

* Last point missing in trajectory
* Acceleration not considered when calculating `ds`


klemense1:
replaced by #179

PR #183, opened 2019-12-01
Optimize pygame viewer

tin1254:
increase framerate from 3 FPS to 130 FPS

### DIFF
--- a/examples/od8_const_vel_one_agent.py
+++ b/examples/od8_const_vel_one_agent.py
@@ -71,8 +71,8 @@ viewer = Panda3dViewer(params=param_server,
                       y_range=[-150, 150],
                       follow_agent_id=agent.id)
 """
-viewer = MPViewer(params=param_server,
-                  follow_agent_id=agent.id)
+# viewer = MPViewer(params=param_server,
+#                   follow_agent_id=agent.id)
 
 # World Simulation
 sim_step_time = param_server["simulation"]["step_time",
--- a/examples/od8_const_vel_one_agent.py
+++ b/examples/od8_const_vel_one_agent.py
@@ -71,8 +71,8 @@ viewer = Panda3dViewer(params=param_server,
                       y_range=[-150, 150],
                       follow_agent_id=agent.id)
 """
-# viewer = MPViewer(params=param_server,
-#                   follow_agent_id=agent.id)
+viewer = MPViewer(params=param_server,
+                  follow_agent_id=agent.id)
 
 # World Simulation
 sim_step_time = param_server["simulation"]["step_time",
--- a/modules/runtime/viewer/pygame_viewer.py
+++ b/modules/runtime/viewer/pygame_viewer.py
@@ -7,6 +7,7 @@ import pygame as pg
 import numpy as np
 from bark.geometry import *
 from bark.viewer import *
+from bark.world.opendrive import *
 from bark.models.dynamic import *
 from modules.runtime.viewer.viewer import BaseViewer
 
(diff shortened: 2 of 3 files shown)


examples/od8_const_vel_one_agent.py
```diff
@@ -71,8 +71,8 @@
                       y_range=[-150, 150],
                       follow_agent_id=agent.id)
 """
-viewer = MPViewer(params=param_server,
-                  follow_agent_id=agent.id)
+# viewer = MPViewer(params=param_server,
```
klemense1: I would not change the default viewer here.

modules/runtime/viewer/pygame_viewer.py
```diff
@@ -32,36 +39,55 @@ def __init__(self, params=None, **kwargs):
             self.screen = None
             print("No available video device")
 
+    def drawMap(self, map):
```
klemense1: drawMap is already defined in the base viewer (modules/runtime/viewer/viewer.py) and thus should not be defined here. What was your motivation here?

modules/runtime/viewer/pygame_viewer.py
```diff
@@ -32,36 +39,55 @@ def __init__(self, params=None, **kwargs):
...
-        line_transformed_np = np.apply_along_axis(self.pointToPG, axis=1, arr=line2d_np)
-        for line_slice in np.ma.clump_unmasked(np.ma.masked_invalid(line_transformed_np)):
-            line_np = line_transformed_np[line_slice]
-            if len(line_np.tolist()) < 2:
-                continue
-            pg.draw.lines(
-                self.screen, self.getColor(color), False,
-                line_np.tolist(), 3)
+        line2d = self.pointsToCamera(line2d)
+        pg.draw.lines(self.screen, self.getColor(color), False, line2d, 3)
```
klemense1: can you elaborate on this chaange? why don't you need the transformation?

modules/runtime/viewer/pygame_viewer.py
tin1254: I saw the map was being drawn again and again, which is not necessary and is one of the two main reasons for the performance drop.
To avoid redrawing, we need to have a "surface" drawn with the map, and "blit" back to the main screen during each running frame.

https://github.com/bark-simulator/bark/blob/7bc54de159edf12368e0994be1a481d743fccfa5/modules/runtime/viewer/pygame_viewer.py#L68
https://github.com/bark-simulator/bark/blob/7bc54de159edf12368e0994be1a481d743fccfa5/modules/runtime/viewer/pygame_viewer.py#L72
I think avoiding redrawing cannot be done without these two lines, and cannot be done with drawline2d, that's why I defined a new drawmap

modules/runtime/viewer/pygame_viewer.py
tin1254: Another main performance drop came from this implementation of transformation `np.apply_along_axis(self.pointToPG, axis=1, arr=...)` and the function calls following by this.
So I decide to using native numpy elementwise calculations to do the transformation with these two function
https://github.com/bark-simulator/bark/blob/7bc54de159edf12368e0994be1a481d743fccfa5/modules/runtime/viewer/pygame_viewer.py#L152-L162

examples/od8_const_vel_one_agent.py
tin1254: reverted the change here

modules/runtime/viewer/pygame_viewer.py
klemense1: can't we use the base class method? es for example in https://stackoverflow.com/questions/805066/call-a-parent-classs-method-from-child-class

tin1254: I think the logic of drawMap in pygame_viewer could be too complicated to adapt to the function in the base class... as to not redraw the map, we have to draw to a different source other than the screen, and calculate the coordinates to display in the screen
Or should I remove the feature of not redrawing the map?

PR #185, opened 2019-12-04
Update driving corridor for lane change

AKreutz:
The behavior model of an agent can cause it to leave its driving corridor, for example when changing lane to overtake an obstacle. In order to still be able to plan a trajectory, the driving corridor of the agent should be updated in this case.

With this pull request, World will attempt to find a new driving corridor for an agent that has moved outside of its old one.
Additionally, it is now also possible to find a driving corridor to a final lane that is parallel to the goal lane, using only successor edges of the roadgraph.


AKreutz:
Closing this pull request and reopening it from a branch that is part of the upstream repository (https://github.com/bark-simulator/bark/pull/187).

PR #198, opened 2020-01-04
Clone local map when cloning agents

cirrostratus1:
The local map also needs to be copied if an agent is cloned, since otherwise the driving corridor in the clone still references the original, which could lead to unwanted side effects.


patrickhart:
As the agents are always copied into the ObservedWorld this would slow BARK too much down.

Issue #213, opened 2020-01-10
Error in the length of ref line in two xodr maps


tin1254:
the length of the single ref line is much larger than the total length of ref line which denoted in `<road>`

klemense1:
we also have to make sure to change that map then in the scenario database.

patrickhart:
And in Bark-ML

PR #214, opened 2020-01-10
fix length of ref line in xodr

tin1254:
fix minor error mentioned in https://github.com/bark-simulator/bark/issues/213


tin1254:
there are also some other errors inside the map, I'll reopen after a thorough check

PR #322, opened 2020-03-31
add sequential goal

ChenyangTang1:


modules/runtime/scenario/scenario_generation/config_with_ease.py
```diff
@@ -135,11 +135,16 @@ def goal(self, world):
     """
     # should be access safe, otherwise would not reach.
     road_corr = world.map.GetRoadCorridor(self._road_ids, XodrDrivingDirection.forward)
-    lane_corr = road_corr.lane_corridors[self._lane_corridor_id]
```
patrickhart: This should not be modified. You should only add the SequentialGoalDefinition in the BARK-ML scenario generation

PR #324, opened 2020-04-08
integrate carla interface

tin1254:
Add functions in bark to access and modify agents state for cosimulation with carla
update and optimize pygame_viewer


modules/runtime/viewer/pygame_viewer.py
```diff
@@ -57,85 +62,177 @@ def drawMap(self, map):
...
+        background_color = kwargs.pop(
+            "background_color", self.background_color)
+        text_surf = pg.font.SysFont(font, font_size).render(
+            text, True, color, background_color)
+        self.screen_surface.blit(
+            text_surf, (position[0] * self.screen_width, (1 - position[1]) * self.screen_height))
 
     def getColor(self, color):
-        if isinstance(color, Viewer.Color):
+        if isinstance(color, str):
```
klemense1, line 190: why did you make these changes regarding color?

modules/world/objects/agent.cpp
```diff
@@ -80,6 +80,16 @@ void Agent::ExecutionPlan(const float &dt) {
                             history_.back().first);
 }
 
+void Agent::addTrajectoryStep(const StateActionPair& state_action_pair){
```
klemense1, line 83: can you elaborate on this function?

modules/world/world.cpp
```diff
@@ -220,7 +220,59 @@ FrontRearAgents World::GetAgentFrontRearForId(
...
+    if (agent) {
+      //! clone current world
+      ObservedWorld observed_world(current_world,
+                                   agent_id);
+      agent->BehaviorPlan(delta_time, observed_world);
+      agent->ExecutionPlan(delta_time);
+
+      trajectory_map[agent_id]=agent->GetExecutionTrajectory();
+    } else {
+      std::cout << "Agent" << agent_id << " doesn't exist." << std::endl;
```
klemense1, line 269: use logging instead of std::cout

modules/runtime/viewer/pygame_viewer.py
tin1254: It was a hard fix to the original code, it could never hit the if case with Viewer.Color, and raise error because once into the else.
For example, somewhere in bark set the color as string "blue".

tin1254: Should I look into the issue?

modules/world/objects/agent.cpp
tin1254: This function is intended to add the state retrieved from Carla to the state history matrix. And in the behavior part, the last state of an agent is retrieved as the current state for planning, I found similar implementation in `Agent::Execute`.

PR #343, opened 2020-05-19
Primitive preconditions

cirrostratus1:
Preconditions for macro action motion primitives


modules/models/dynamic/single_track.hpp
```diff
@@ -16,22 +16,22 @@ namespace modules {
...
-                        "Maximum lateral acceleration [m/s^2]", 4.0);
-  }
+        wheel_base_(params->GetReal("DynamicModel::wheel_base",
+                                    "Wheel base of vehicle [m]", 2.7)),
+        steering_angle_max_(params->GetReal(
+            "DynamicModel::delta_max", "Maximum Steering Angle [rad]", 0.2)),
+        lat_acceleration_max_(
+            params->GetReal("DynamicModel::lat_acc_max",
+                            "Maximum lateral acceleration [m/s^2]", 4.0)),
+        max_acceleration_(params->GetReal("DynamicModel::max_acceleration",
```
klemense1, line 30: In accordance with lat_acceleration_max_, we should name this long_acceleration_max_ (similar with _min). Also consider removing the dynamic / static description.

modules/models/tests/behavior_motion_primitive_test.cc
```diff
@@ -101,41 +123,79 @@ TEST(behavior_motion_primitives_plan, behavior_test) {
...
+    LOG(WARNING) << a.first << ", ";
+  }
+  auto observed_worlds = world->Observe({1, 4});
+  auto corridors0 = GetCorridors(observed_worlds[0]);
+  EXPECT_TRUE(
+      primitive.IsPreConditionSatisfied(observed_worlds[0], corridors0));
+  auto corridors1 = GetCorridors(observed_worlds[1]);
+  EXPECT_FALSE(
+      primitive.IsPreConditionSatisfied(observed_worlds[1], corridors1));
   // auto traj = primitive.Plan(0.5, world1);
```
klemense1, line 185: What is tested here is the precondition. Can we test for the action as well? Otherwise, rename the test function

PR #364, opened 2020-06-05
Changed World, Agent and the Interpolation Model

patrickhart:
Behavior models can now have an arbitrary planning horizon and time discretization.
The Execute function has been moved from the agent to the world.
Additionally, execution models now also can set a valid or invalid flag.

An assert has been introduced in the World::Execute(..) function to make sure all agents have the same world time.


docs/source/models.md
```diff
@@ -28,7 +28,7 @@ class BehaviorModel : public modules::commons::BaseType {
         last_action_(),
         behavior_status_(status) {}
   ...
-  virtual Trajectory Plan(float delta_time,
+  virtual Trajectory Plan(float min_dt,
```
juloberno, line 31: may be call min_planning_time ?

modules/models/behavior/behavior_model.hpp
```diff
@@ -73,19 +77,26 @@ class BehaviorModel : public modules::commons::BaseType {
...
+  Action GetAction() const { return action_to_behavior_; }
+  virtual void ActionToBehavior(const Action& action) {
+    action_to_behavior_ = action;
+  };
 
  private:
   dynamic::Trajectory last_trajectory_;
+  // can either be the last action or action to be executed
   Action last_action_;
+  Action action_to_behavior_;
```
juloberno, line 99: Do we need to support the action to behavior for all behavior models? I think it is sufficient for motion primitives or these types of behavior models

modules/models/execution/interpolation/interpolate.cpp
```diff
@@ -13,16 +13,115 @@ namespace modules {
...
+    if (world_time >= trajectory(i, TIME_POSITION)) {
+      ret_idx = i;
+      found_closest_pt = true;
+      break;
+    }
+  }
+  return {ret_idx, found_closest_pt};
+}
+
+State ExecutionModelInterpolate::Interpoalte(
```
juloberno, line 75: typo

modules/models/execution/execution_model.hpp
```diff
@@ -22,33 +22,58 @@ using dynamic::DynamicModelPtr;
...
   ExecutionModel(const ExecutionModel& execution_model) :
     BaseType(execution_model.GetParams()),
-    last_trajectory_(execution_model.GetLastTrajectory()) {}
+    last_state_(execution_model.GetLastState()),
+    last_trajectory_(execution_model.GetLastTrajectory()),
+    execution_status_(execution_model.GetExecutionStatus()) {}
 
   virtual ~ExecutionModel() {}
 
+  State GetLastState() const { return last_state_; }
```
juloberno, line 48: maybe call this GetExecutedState() ? last state could be unclear

modules/world/world.cpp
```diff
@@ -41,6 +43,50 @@ World::World(const std::shared_ptr<World>& world) :
...
+void World::Execute(const float& world_time) {
+  using models::dynamic::StateDefinition::TIME_POSITION;
+  for (auto agent : agents_) {
+    if (agent.second->GetBehaviorStatus() == BehaviorStatus::VALID &&
+        agent.second->GetExecutionStatus() == ExecutionStatus::VALID) {
+      agent.second->UpdateStateAction();
+      // make sure all agents have the same world time
+      // otherwise the simulation is not correct
+      const auto& agent_state = agent.second->GetCurrentState();
+      assert(fabs(agent_state(TIME_POSITION) - world_time) < 0.01);
```
juloberno, line 74: use BARK_EXPECT_TRUE here

modules/world/world.cpp
```diff
@@ -41,6 +43,50 @@ World::World(const std::shared_ptr<World>& world) :
...
+  }
+  RemoveInvalidAgents();
+}
+
+WorldPtr World::GetWorldAtTime(const float& world_time) const {
+  WorldPtr current_world_state(this->Clone());
+  for (auto agent : current_world_state->GetAgents()) {
+    if (agent.second->GetBehaviorStatus() == BehaviorStatus::VALID)
+      agent.second->PlanExecution(world_time);
+    agent.second->UpdateStateAction();
```
juloberno, line 85: if Execution Valid should be checked here

modules/models/behavior/behavior_model.hpp
patrickhart: Due to the fact that is is wrapped in the PyBehaviorModel I think there is no other option?

docs/source/models.md
patrickhart: Will change that!

modules/models/execution/interpolation/interpolate.cpp
patrickhart: Ups.

modules/models/execution/execution_model.hpp
patrickhart: sounds good!

modules/world/world.cpp
patrickhart: okay!

modules/world/world.cpp
patrickhart: I do not think so, as PlanExecution calls the exxecution model of the agent.

PR #372, opened 2020-06-17
Benchmark graceful degredation

cirrostratus1:
Error handling for benchmark execution


modules/benchmark/benchmark_runner.py
```diff
@@ -140,7 +162,7 @@ def _create_configurations(self, num_scenarios=None):
                     benchmark_configs.append(benchmark_config)
         return benchmark_configs
 
-    def run(self, viewer=None, maintain_history=False):
+    def run(self, viewer=None, maintain_history=False, stage_dir=None):
```
patrickhart, line 165: what does stage_dir imply?

cirrostratus1: It defines the directory where intermediate results (stages) are stored.

cirrostratus1: https://github.com/bark-simulator/bark/blob/7a21ff1ccf2e430ee3756c13708df2562d7d5593/modules/benchmark/benchmark_runner.py#L178-L188

modules/benchmark/benchmark_runner.py
```diff
@@ -118,14 +140,14 @@ def __init__(self,
...
     def _create_configurations(self, num_scenarios=None):
         benchmark_configs = []
-        for behavior_name, behavior_bark in self.behaviors.items():
-            # run over all scenario generators from benchmark database
-            for scenario_generator, scenario_set_name in self.benchmark_database:
-                for scenario, scenario_idx in scenario_generator:
+        # run over all scenario generators from benchmark database
+        for scenario_generator, scenario_set_name in self.benchmark_database:
+            for scenario, scenario_idx in scenario_generator:
+                for behavior_name, behavior_bark in self.behaviors.items():
```
klemense1, line 150: is there any reason you changed the order of the loops?

cirrostratus1: The previous order of the loops generated a sequence of configurations so that all scenarios would be run for on behavior before any other behavior is tested.

1) Scenario A, Behavior A
2) Scenario B, Behavior A
3) Scenario A, Behavior B
4) Scenario B, Behavior B

 When saving intermediate results, it appeared more useful first to complete a scenario with all behaviors, than running all scenarios on the same behavior. The intermediate result should then still be meaningful even though not all scenarios have been tested yet, compared to the previous order, where we would only have the results for one single behavior.

1) Scenario A, Behavior A
2) Scenario A, Behavior B
3) Scenario B, Behavior A
4) Scenario B, Behavior B

modules/benchmark/benchmark_runner.py
patrickhart: Okay, just talked to Klemens about that. Seems fine for me.

modules/benchmark/benchmark_runner.py
patrickhart: I guess the order does not really matter for the pandas table. Thus, should be okay.

Issue #371, opened 2020-06-16
PIP package for BARK


patrickhart:
Build a pip package for BARK to make it more accessible.

patrickhart:
#373

PR #373, opened 2020-06-17
Pip package for BARK

mansoorcheema:
Implemented pip packaging for #371


WORKSPACE
```diff
@@ -28,7 +28,7 @@ planner_uct_rules_dependencies()
 # -------- Benchmark Database -----------------------
 git_repository(
   name = "benchmark_database",
-  commit="a5d97f80eb7d6d0fe1d4221c9bc3694e816f1dfe",
+  commit="ca6f2db654bb3945a7c23ab7f04268e7d68fa0c5",
```
patrickhart, line 31: Add a pull request for the benchmark database, so we can link to a sha on the master.

bark/benchmark/tests/py_benchmark_process_tests.py
```diff
@@ -10,20 +10,22 @@
...
 
-from bark.world.evaluation import *
-from modules.runtime.commons.parameters import ParameterServer
-from bark.models.behavior import BehaviorIDMClassic, BehaviorConstantVelocity
+from bark.core.world.evaluation import *
+from bark.runtime.commons.parameters import ParameterServer
+from bark.core.models.behavior import BehaviorIDMClassic, BehaviorConstantVelocity
 
 # to find database files
+print ('currdir:',os.getcwd())
```
patrickhart, line 26: could remove this log messages

mansoorcheema:
@patrickhart @juloberno Could you provide your PyPi username so i could add you as maintainer.

patrickhart:
My PyPI username is "patrickhrt". Thanks!

docs/source/guidelines.md
```diff
@@ -14,7 +14,7 @@ You now need to define the path to cpplint as local user.
 ```
 Cpplint: Cpplint Path
 The path to the cpplint executable. If not set, the default location will be used.
-/(YOUR LOCAL PATH FOR BARK)/bark/python/venv/bin/cpplint
+/(YOUR LOCAL PATH FOR BARK)/bark/python_wrapper/venv/bin/cpplint
```
klemense1, line 17: shouldn't we install the python env in enother dir different from python_wrapper?

patrickhart:
Should/can we use absolute paths in the Readme.md?

At the moment the images are not shown here: https://pypi.org/project/bark-simulator/

docs/source/guidelines.md
mansoorcheema: Its already in python folder. I mistakenly updated this document. Have pushed the correction.

mansoorcheema:
@patrickhart Updated readme.md with absolute path. The images work fine in Pypi and Github.

patrickhart:
Another question would be what we can run when we checkout the pip-package. We cannot run the examples anyway, right? So would it not be better to move these out of the bark folder to the highest hierarchy to highlight the example folder more?

What is your opinion on that @mansoorcheema?

PR #377, opened 2020-06-18
Evaluator for LTL Rules

cirrostratus1:


modules/runtime/scenario/scenario.py
```diff
@@ -72,7 +72,7 @@ def _build_world_state(self):
     for agent in self._agent_list:
       agent.GenerateRoadCorridor(self._map_interface)
       world.AddAgent(agent)
-    world.UpdateAgentRTree()
+      world.UpdateAgentRTree()
```
patrickhart, line 75: why update it every single time? Please provide an explanation.

modules/runtime/scenario/scenario_generation/config_readers/rule_generation.py
```diff
@@ -0,0 +1,72 @@
...
+
+import logging, sys
+from bark.ltl import RuleMonitor
+from bark.world.evaluation import *
+from bark.world.opendrive import XodrLaneType, XodrRoadMark, XodrRoadMarkType
+from bark.geometry import Point2d
+from modules.runtime.commons.parameters import ParameterServer
+
+
+def make_rules(rule_params):
```
patrickhart, line 14: MakeRules(..)

modules/runtime/scenario/scenario_generation/config_readers/rule_generation.py
```diff
@@ -0,0 +1,72 @@
...
+        param_srv.convert_to_param(rule_json)
+        rules.append(RuleMonitor(param_srv["formula", "Traffic rules described using LTL"],
+                                      param_srv["weight", "Penalty for violating the rule.", 0.0],
+                                      param_srv[
+                                          "priority", "Index of the entry in the reward vector to add penalty.", 0]))
+
+    return rules
+
+
+def dump_rules(rules):
```
patrickhart, line 27: DumpRules(..)

modules/runtime/scenario/scenario_generation/config_readers/rule_generation.py
```diff
@@ -0,0 +1,72 @@
...
+    return AgentBeyondPointLabelFunction(label_str, Point2d(point[0], point[1]))
+
+
+def make_EgoBeyondPointLabelFunction(params):
+    label_str = params["label_str"]
+    point = params["point"]
+    return EgoBeyondPointLabelFunction(label_str, Point2d(point[0], point[1]))
+
+
+def make_SafeDistanceLabelFunction(params):
```
patrickhart, line 49: make function naming congruent

modules/runtime/scenario/scenario_generation/interaction_dataset_scenario_generation.py
```diff
@@ -22,6 +22,7 @@
 class InteractionDatasetScenarioGeneration(ScenarioGeneration):
 
     def __init__(self, params=None, num_scenarios=None, random_seed=None):
+        self.interaction_ds_reader = InteractionDatasetReader()
```
patrickhart, line 25: use underscore

modules/runtime/scenario/scenario_generation/interaction_dataset_scenario_generation_full.py
```diff
@@ -4,23 +4,28 @@
...
 
 class InteractionDatasetScenarioGenerationFull(ScenarioGeneration):
-  # This class reads in a track file from the interaction dataset
-  # and generates a scenario for each agent as the eval agent.
+    # This class reads in a track file from the interaction dataset
+    # and generates a scenario for each agent as the eval agent.
 
     def __init__(self, params=None, num_scenarios=None, random_seed=None):
+        self._map_interface = None
+        self.interaction_ds_reader = InteractionDatasetReader()
```
patrickhart, line 28: use underscore

modules/runtime/tests/py_scenario_generation_tests.py
```diff
@@ -142,10 +142,12 @@ def test_dataset_scenario_generation_full(self):
         track_filename = "modules/runtime/tests/data/interaction_dataset_dummy_track.csv"
 
         params["Scenario"]["Generation"]["InteractionDatasetScenarioGenerationFull"]["MapFilename"] = map_filename
-        params["Scenario"]["Generation"]["InteractionDatasetScenarioGenerationFull"]["TrackFilename"] = track_filename
+        params["Scenario"]["Generation"]["InteractionDatasetScenarioGenerationFull"]["TrackFilenameList"] = [track_filename]
 
         scenario_generation = InteractionDatasetScenarioGenerationFull(
             params=params, num_scenarios=2)
+          
+        assert(scenario_generation.get_num_scenarios() == 2)
```
patrickhart, line 150: assert needs to be without brackets?

modules/world/evaluation/evaluator_being_overtaken.hpp
```diff
@@ -0,0 +1,46 @@
...
+ public:
+  explicit EvaluatorBeingOvertakenAssumption(AgentId agent_id)
+      : EvaluatorLTL(agent_id, formula_) {}
+  EvaluationReturn Evaluate(const world::World& world) override {
+    auto cloned_world = world.Clone();
+    cloned_world->AddLabels(labels_);
+    return EvaluatorLTL::Evaluate(*cloned_world);
+  }
+
+  static const char formula_[];
```
patrickhart, line 39: why does this need to be static?

modules/world/evaluation/evaluator_being_overtaken.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "modules/world/evaluation/evaluator_being_overtaken.hpp"
+
+#include "modules/world/evaluation/labels/agent_near_label_function.hpp"
+#include "modules/world/evaluation/labels/ego_accelerate_label_function.hpp"
+#include "modules/world/evaluation/labels/right_of_label_function.hpp"
+
+namespace modules {
+namespace world {
+namespace evaluation {
+const char EvaluatorBeingOvertaken::formula_[] =
```
patrickhart, line 15: In general, please provide docstrings to foster a general understanding.

modules/world/evaluation/evaluator_ltl.cpp
```diff
@@ -0,0 +1,130 @@
...
+    // Add new agents to known agents
+    AddKnownAgents(current_agents);
+    // Remove rule states for non-existent agents
+    RemoveRuleStates(current_agents);
+  }
+
+  // Obtain label values
+  auto labels = observed_world.EvaluateLabels();
+  unsigned int guarantee_violations = 0;
+  float penalty;
```
patrickhart, line 65: initialize float value

modules/world/evaluation/evaluator_right_overtake.cpp
```diff
@@ -0,0 +1,32 @@
...
+#include "modules/world/evaluation/labels/behind_of_label_function.hpp"
+#include "modules/world/evaluation/labels/dense_traffic_label_function.hpp"
+#include "modules/world/evaluation/labels/front_of_label_function.hpp"
+#include "modules/world/evaluation/labels/right_of_label_function.hpp"
+
+namespace modules {
+namespace world {
+namespace evaluation {
+const char EvaluatorRightOvertake::formula_[] =
+    "G (!dense -> !(b_v#0 & X[!](b_v#0 U r_v#0 U f_v#0)))";
```
patrickhart, line 17: add what this means

modules/world/evaluation/evaluator_speed_advantage_overtake.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "modules/world/evaluation/labels/behind_of_label_function.hpp"
+#include "modules/world/evaluation/labels/front_of_label_function.hpp"
+#include "modules/world/evaluation/labels/left_of_label_function.hpp"
+#include "modules/world/evaluation/labels/rel_speed_label_function.hpp"
+#include "modules/world/evaluation/labels/right_of_label_function.hpp"
+
+namespace modules {
+namespace world {
+namespace evaluation {
+const char EvaluatorSpeedAdvantageOvertake::formula_[] =
```
patrickhart, line 17: add description or link to paper

modules/world/evaluation/evaluator_zip_merge_chn.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "modules/world/evaluation/labels/agent_near_label_function.hpp"
+#include "modules/world/evaluation/labels/behind_of_label_function.hpp"
+#include "modules/world/evaluation/labels/ego_beyond_point_label_function.hpp"
+#include "modules/world/evaluation/labels/left_of_label_function.hpp"
+#include "modules/world/evaluation/labels/preceding_agent_label_function.hpp"
+
+namespace modules {
+namespace world {
+namespace evaluation {
+const char EvaluatorZipMergeChn::formula_[] =
```
patrickhart, line 19: add descr.

modules/world/evaluation/evaluator_zip_merge_deu.cpp
```diff
@@ -0,0 +1,37 @@
...
+#include "modules/world/evaluation/labels/agent_near_label_function.hpp"
+#include "modules/world/evaluation/labels/behind_of_label_function.hpp"
+#include "modules/world/evaluation/labels/ego_beyond_point_label_function.hpp"
+#include "modules/world/evaluation/labels/left_of_label_function.hpp"
+#include "modules/world/evaluation/labels/preceding_agent_label_function.hpp"
+
+namespace modules {
+namespace world {
+namespace evaluation {
+const char EvaluatorZipMergeDeu::formula_[] =
```
patrickhart, line 19: add descr.

modules/world/evaluation/labels/base_label_function.hpp
```diff
@@ -0,0 +1,48 @@
...
+using ltl::Label;
+using LabelMap = ltl::EvaluationMap;
+
+class BaseLabelFunction {
+ public:
+  explicit BaseLabelFunction(std::string label_str)
+      : label_str_(std::move(label_str)) {}
+  virtual std::vector<LabelMap::value_type> Evaluate(
+      const world::ObservedWorld &observed_world) const = 0;
+  const std::string &GetLabelStr() const { return label_str_; }
```
patrickhart, line 34: reformat so std::string& everywhere

modules/world/tests/label_test.cc
```diff
@@ -0,0 +1,223 @@
+// Copyright (c) 2020 fortiss GmbH
```
patrickhart, line 1: should we split up the world tests as well in subfolders?

modules/world/world.cpp
```diff
@@ -123,7 +123,7 @@ EvaluationMap World::Evaluate() const {
 }
 
 std::vector<ObservedWorld> World::Observe(
-    const std::vector<AgentId>& agent_ids) {
+    const std::vector<AgentId>& agent_ids) const {
```
patrickhart, line 126: Do not use const here. otherwise e.g. a PredictionSetting cannot be set up.
The behavior of other vehicles should still be changable.

modules/world/world.cpp
```diff
@@ -268,6 +268,18 @@ FrontRearAgents World::GetAgentFrontRearForId(
   return fr_agents;
 }
 
+void World::AddLabels(const LabelFunctions& label_evaluators) {
```
patrickhart, line 271: Why do we add the Labels to the World and not to the Evaluator?
It clutters the world.

modules/world/world.hpp
```diff
@@ -94,7 +99,8 @@ class World : public commons::BaseType {
   /**
    * @brief  Generates and ObservedWorld for the specified agents
    */
-  std::vector<ObservedWorld> Observe(const std::vector<AgentId>& agent_ids);
+  std::vector<ObservedWorld> Observe(
```
patrickhart, line 102: as mentioned before, I do not think this should be const.

modules/world/world.hpp
```diff
@@ -159,9 +167,14 @@ class World : public commons::BaseType {
     ClearObjects();
     ClearEvaluators();
   }
+  void AddLabels(const LabelFunctions& label_evaluators);
```
patrickhart, line 170: Labels should be contained in the evaluator.
If these are required for the LTL rule evaluator, consider to move them there.

python/world/world.cpp
```diff
@@ -40,13 +40,16 @@ void python_world(py::module m) {
...
+    .def("RemoveAgentById", &World::RemoveAgentById)
     .def("AddObject", &World::AddObject)
     .def("GetParams", &World::GetParams)
     .def("UpdateAgentRTree", &World::UpdateAgentRTree)
     .def("ClearEvaluators", &World::ClearEvaluators)
     .def("SetMap", &World::SetMap)
     .def("AddEvaluator", &World::AddEvaluator)
-    .def("GetNearestAgents", &World::GetNearestAgents)
+    .def("AddLabels", &World::AddLabels)
+      .def_property_readonly("labels", &World::GetLabelFunctions)
```
patrickhart, line 51: format

modules/runtime/scenario/scenario.py
cirrostratus1: I think this was unintended. Sorry for that.

python/world/world.cpp
cirrostratus1: This kind of problem happens often to me. I use the CLang Formatter with the format style file of Bark. When running the formatter on the file, it indents the lines by 4 spaces, as happened here. I have the impression that the formatter you use only has a 2 space indent, but the .clang-format file says 4 spaces for continuation.

https://github.com/bark-simulator/bark/blob/cc393384134188bf68d098a9efa486ff36069577/.clang-format#L52

Could you clarify this?

modules/world/world.cpp
cirrostratus1: Even though not implemented yet, the idea was that labels only have to be evaluated once for all evaluators, since some LTL rules are sharing the same labels. This should reduce the execution time, as labels do not have to be reevaluated. For this reason we would not want to store the labels in the evaluators. Another reason is, that LabelFunctions would also be directly passed to successor states of the world. This mechanism is currently used by the mvmcts-planner.

modules/models/behavior/static_trajectory/behavior_static_trajectory.cpp
```diff
@@ -61,7 +61,9 @@ std::pair<int, int> BehaviorStaticTrajectory::Interpolate(
   StateRowVector delta;
   double alpha;
   int idx = -1;
-  BARK_EXPECT_TRUE(static_trajectory_.rows() > 1);
+  if (static_trajectory_.rows() < 2) {
+    return {-1, -1};
+  }
```
klemense1, line 66: was this necessary since we introduced the expired status flag? This might make errors in the dataset harder to debug?

modules/runtime/scenario/scenario.py
klemense1: Please provide a fix for that.

modules/world/world.cpp
patrickhart: however, if used and only used by the mvmcts planner, this seems to be a planner specific feature that is otherwise not needed.

I like the work you have done with the ltl-evaluators. However, I still do not see the necessity of these to be in the world.

I think this is a minor speed draw-back vs. a more cluttered implementation and more cloning effort (cloning the world in every step has an overhead on its own)

python/world/world.cpp
patrickhart: I am currently working under macOS and do not have this problem.
Maybe @klemense1 could shed some light on this issue?

modules/world/evaluation/evaluator_right_overtake.cpp
klemense1: rename predicates so they are congruent with the conventions in the other labels

modules/world/evaluation/evaluator_speed_advantage_overtake.cpp
klemense1: we will provide a link once the paper got accepted. add some small doc string, then it should be okay for now.

modules/world/evaluation/evaluator_zip_merge_chn.cpp
klemense1: we will provide a link once the paper got accepted. add some small doc string, then it should be okay for now.

modules/world/evaluation/evaluator_zip_merge_deu.cpp
klemense1: same as with other evaluators

modules/world/evaluation/labels/ego_accelerate_label_function.cpp
```diff
@@ -0,0 +1,39 @@
...
+      acc_thres_(acc_thres) {}
+
+std::vector<LabelMap::value_type> EgoAccelerateLabelFunction::Evaluate(
+    const world::ObservedWorld& observed_world) const {
+  bool accel = false;
+  const auto ego = observed_world.GetEgoAgent();
+  const auto& history = ego->GetStateInputHistory();
+  if (history.size() > 2) {
+    const auto dx =
+        history.back().first - (history.end() - 2)->first;
```
klemense1, line 29: why use history.back() once and history.end() the other time?

modules/world/world.cpp
patrickhart: And if a label describes the world state. Should this not be done maybe relative to a single vehicle. The the observed world would be a better place to add these labels.

modules/world/evaluation/labels/safe_distance_label_function.cpp
```diff
@@ -0,0 +1,124 @@
...
+      to_rear_(to_rear),
+      delta_(delta),
+      a_e_(a_e),
+      a_o_(a_o) {}
+
+std::vector<LabelMap::value_type> SafeDistanceLabelFunction::Evaluate(
+    const world::ObservedWorld& observed_world) const {
+  auto ego = std::const_pointer_cast<Agent>(observed_world.GetEgoAgent());
+  auto lane_corridor =
+      ego->GetRoadCorridor()->GetNearestLaneCorridor(ego->GetCurrentPosition());
```
klemense1, line 28: why can't we use GetCurrentLaneCorridor() here?

modules/world/tests/label_test.cc
cirrostratus1: Do you mean like `modules/world/evaluation/labels/tests/label_test.cc` ?

modules/world/evaluation/labels/agent_at_lane_end_label_function.cpp
```diff
@@ -0,0 +1,32 @@
...
+  assert(distance_thres_ >= 0.0);
+}
+
+bool modules::world::evaluation::AgentAtLaneEndLabelFunction::EvaluateAgent(
+    const modules::world::ObservedWorld &observed_world,
+    const AgentPtr &other_agent) const {
+  if (other_agent) {
+    const auto &other_pos = other_agent->GetCurrentPosition();
+    const auto &lc =
+        other_agent->GetRoadCorridor()->GetNearestLaneCorridor(other_pos);
```
klemense1, line 26: why can't we use GetCurrentLaneCorridor() here?

modules/world/map/road_corridor.cpp
```diff
@@ -37,6 +38,22 @@ RoadCorridor::GetLeftRightLaneCorridor(const Point2d& pt) const {
     GetLaneCorridor(right_lane_id));
 }
 
+LaneCorridorPtr RoadCorridor::GetNearestLaneCorridor(const Point2d& pt) const {
```
klemense1, line 41: do we really need this? I assume this to be slower than GetCurrentLaneCorridor()

modules/world/tests/label_test.cc
patrickhart: for example, or just split the test folder within world/tests/ as the number of tests is steadily increasing.

modules/world/evaluation/evaluator_being_overtaken.hpp
cirrostratus1: The formula is specific for the purpose of the evaluator. Thus it is identical for every instance of the evaluator and only needs to be stored once.

modules/world/map/road_corridor.cpp
cirrostratus1: In principal yes, but for example when working on the INTERACTION dataset, the reference points of the agents are sometimes slightly out of the lane corridor. `GetNearestLaneCorridor` first calls `GetCurrentLaneCorridor` and if a null pointer is returned, it searches for the nearest lane. So it should be only slower, in the exceptional case that an agent can not be assigned to a lane corridor.

modules/world/evaluation/labels/agent_at_lane_end_label_function.cpp
cirrostratus1: See https://github.com/bark-simulator/bark/pull/377#discussion_r442195201

modules/world/evaluation/labels/safe_distance_label_function.cpp
cirrostratus1: See https://github.com/bark-simulator/bark/pull/377#discussion_r442195201

modules/world/map/road_corridor.cpp
klemense1: okay

modules/models/behavior/static_trajectory/behavior_static_trajectory.cpp
cirrostratus1: Either we have to somehow handle this in the behavior model or we have to skip agents in the scenario generation that only have one trajectory point.

modules/world/evaluation/labels/base_label_function.hpp
cirrostratus1: I suggest to change

https://github.com/bark-simulator/bark/blob/cc393384134188bf68d098a9efa486ff36069577/.clang-format#L54

to `DerivePointerAlignment: false` then.

python/world/world.cpp
klemense1: Luis is right, the indentation here is not correct according to the clang file.

modules/world/evaluation/labels/base_label_function.hpp
klemense1: I changed it to false

bark/runtime/scenario/interaction_dataset_processing/__init__.py
```diff
@@ -1,19 +1,21 @@
...
-from .interaction_dataset_reader import behavior_from_track
-from .interaction_dataset_reader import track_from_trackfile
-from .interaction_dataset_reader import agent_from_trackfile
+from .interaction_dataset_reader import BarkStateFromMotionState
+from .interaction_dataset_reader import TrajectoryFromTrack
+from .interaction_dataset_reader import ShapeFromTrack
+from .interaction_dataset_reader import InitStateFromTrack
+from .interaction_dataset_reader import GoalDefinitionFromTrack
+from .interaction_dataset_reader import BehaviorFromTrack
+#from .interaction_dataset_reader import TrackFromTrackfile
```
cirrostratus1, line 7: I think we can remove this.

PR #394, opened 2020-07-03
LTL Evaluator Fixes

klemense1:
- importing LTL evaluators in benchmark runner
- bug fixes in rule and label

### DIFF
--- a/bark/benchmark/benchmark_runner.py
+++ b/bark/benchmark/benchmark_runner.py
@@ -18,6 +18,11 @@ from bark.runtime.commons.parameters import ParameterServer
 from bark.runtime.scenario.scenario import Scenario
 from bark.core.world.evaluation import *
 
+# ltl lib is only build conditionally
+try:
+  from bark.core.world.evaluation.ltl import *
+except:
+  pass
 
 # contains information for a single benchmark run
 class BenchmarkConfig:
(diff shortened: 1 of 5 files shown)


bark/python_wrapper/BUILD
```diff
@@ -9,8 +9,7 @@ cc_binary(
     "world/*.cpp", "world/*.hpp",
     "models/*.cpp", "models/*.hpp",
     "viewer/*.cpp", "viewer/*.hpp",
-    "runtime.hpp"], exclude=["world/ltl.cpp", "world/ltl.hpp"])
-    + select({"//bark/world/evaluation/ltl:_ltl_rules": ["world/ltl.cpp", "world/ltl.hpp"], "//conditions:default": []}),
```
juloberno: why is it non default? Build time?

bark/benchmark/benchmark_runner.py
```diff
@@ -18,6 +18,11 @@
 from bark.runtime.scenario.scenario import Scenario
 from bark.core.world.evaluation import *
 
+# ltl lib is only build conditionally
+try:
+  from bark.core.world.evaluation.ltl import *
+except:
+  pass
```
juloberno, line 25: not so nice, to handle conditional deps via try and catch, maybe add another evaluation specification not based on string but on type, so instead of "EvaluatorLTL" one can also use EvaluatorLTL

patrickhart: Could add an wrapper class for the evaluator that pre-defines which LTL Label are used in this one. Otherwise, do the evaluators have to be added here and can’t we pass them via lambda function? Have to agree that a try and except might not be the best solution. But would be acceptable for me.

PR #397, opened 2020-07-11
EvaluatorLTL can be parameterized

klemense1:
- omit other Rule Evaluators, as they can be constructed from EvaluatorLTL using Parameters
- some fixes


bark/benchmark/benchmark_runner_mp.py
```diff
@@ -105,14 +105,15 @@ def __init__(self, benchmark_database=None,
           ray.init(num_cpus=num_cpus, memory=memory_total*0.3, object_store_memory=memory_total*0.7, \
              _internal_config='{"initial_reconstruction_timeout_milliseconds": 100000}') # we split memory between workers (30%) and objects (70%)
         
+        serialized_evaluators = pickle.dumps(evaluators)
         ray.register_custom_serializer(
```
juloberno, line 109: a custom serializer does not work?

klemense1: no, unfortunately not

bark/benchmark/benchmark_runner.py
```diff
@@ -37,7 +38,11 @@ def __init__(self,
                  merge_existing=False):
 
         self.benchmark_database = benchmark_database
-        self.evaluators = evaluators or {}
+        if evaluators is not None and isinstance(evaluators, bytes):
```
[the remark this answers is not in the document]
klemense1: Okay, I did just that, see the new commit.

PR #398, opened 2020-07-11
removing incorrect handling of pickle file

klemense1:
saving and loading a pickle file caused the following error:

```dmp = pickle.load(handle)
_pickle.UnpicklingError: A load persistent id instruction was encountered, but no persistent_load function was specified.
```

removing the special handling of pickle files during loading solves this.


klemense1:
> Why to change to that? I still have some old files using pickle, which then can't be loaded anymore. We should keep the load_pickle function.
> 
> Maybe better to introduce some function parameter, use_pickle_format and add it to both save and load functions.

I added a logging warning message, but reverted the deletion

PR #406, opened 2020-08-14
Increase robustness of GetAgentInFront Function

klemense1:
- if agent's touch the lane corridor (which happens quite often in the dataset), they are not really the agent in front. 
- thus lateral distance to current lane is now taken into account: if it's larger than half the lane width, this agent is not regarded as the agent in front.
- also, some illustration features to labels are added to the viewer (made optional)

### DIFF
--- a/bark/benchmark/benchmark_runner.py
+++ b/bark/benchmark/benchmark_runner.py
@@ -203,7 +203,6 @@ class BenchmarkRunner:
             if not terminal:
                 if viewer:
                     viewer.drawWorld(world, scenario._eval_agent_ids, scenario_idx=benchmark_config.scenario_idx)
-                    viewer.saveFig("/tmp/viewer_{}.png".format(step))
                     viewer.show(block=False)
                     time.sleep(step_time)
                     viewer.clear()
--- a/bark/benchmark/benchmark_runner.py
+++ b/bark/benchmark/benchmark_runner.py
@@ -203,6 +203,7 @@ class BenchmarkRunner:
             if not terminal:
                 if viewer:
                     viewer.drawWorld(world, scenario._eval_agent_ids, scenario_idx=benchmark_config.scenario_idx)
+                    viewer.saveFig("/tmp/viewer_{}.png".format(step))
                     viewer.show(block=False)
                     time.sleep(step_time)
                     viewer.clear()
--- a/bark/geometry/line.hpp
+++ b/bark/geometry/line.hpp
@@ -255,6 +255,10 @@ inline bool CheckSForSegmentIntersection(Line l, float s) {
   return start_it != start_it_low;
 }
 
+inline Point2d GetPointAtIdx(const Line& l, int idx) {
+  return l.obj_.at(idx);
+}
+
 inline Point2d GetPointAtS(Line l, float s) {
   const size_t& length = l.obj_.size();
   if (length <= 1) {  // this is an error Line consist of 0 or 1 element
--- a/bark/geometry/line.hpp
+++ b/bark/geometry/line.hpp
@@ -255,8 +255,13 @@ inline bool CheckSForSegmentIntersection(Line l, float s) {
   return start_it != start_it_low;
 }
 
-inline Point2d GetPointAtIdx(const Line& l, int idx) {
-  return l.obj_.at(idx);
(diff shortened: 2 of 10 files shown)


bark/benchmark/benchmark_runner.py
```diff
@@ -203,6 +203,7 @@ def _run_benchmark_config(self, benchmark_config, viewer=None, maintain_history=
             if not terminal:
                 if viewer:
                     viewer.drawWorld(world, scenario._eval_agent_ids, scenario_idx=benchmark_config.scenario_idx)
+                    viewer.saveFig("/tmp/viewer_{}.png".format(step))
```
patrickhart, line 206: I would suggest to use a file-group and save the images there.
later these still can be accessed in the bazel-bin dir.

bark/geometry/line.hpp
```diff
@@ -255,6 +255,10 @@ inline bool CheckSForSegmentIntersection(Line l, float s) {
   return start_it != start_it_low;
 }
 
+inline Point2d GetPointAtIdx(const Line& l, int idx) {
+  return l.obj_.at(idx);
```
patrickhart, line 259: this is an unsafe access!
check if idx exists, otherwise it might crash.

patrickhart: If idx > l._obj.size() maybe return the last point? otherwise the first? add log message for that as well

bark/runtime/viewer/viewer.py
```diff
@@ -48,6 +49,8 @@ def __init__(self, params=None, **kwargs):
         self.plane_alpha = params["Visualization"]["Map"]["Plane"]["Alpha", "Alpha of the background plane", 1.0]
         self.map_linewidth = params["Visualization"]["Map"]["XodrLanes"]["Boundaries"]["Linewidth", "Linewidth of linestrings", 1.0]
 
+        self.draw_ltl_debug_info= params["Visualization"]["Evaluation"]["DrawLTLDebugInfo", "Flag to specify if debug info to ltl evaluators shall be plotted", False]
```
patrickhart, line 52: formatting

bark/runtime/viewer/viewer.py
```diff
@@ -254,6 +268,19 @@ def drawWorld(self, world, eval_agent_ids=None, filename=None, scenario_idx=None
         if debug_text:
           self.drawText(position=(0.1, 0.9), text="Scenario: {}".format(scenario_idx), fontsize=14)
           self.drawText(position=(0.1, 0.95), text="Time: {:.2f}".format(world.time), fontsize=14)
+
```
patrickhart, line 271: why do we not put all of this in a different function?

bark/world/map/lane_corridor.hpp
```diff
@@ -53,6 +53,15 @@ struct LaneCorridor {
     return GetLength() - GetS(pt);
   }
 
+  float GetLaneWidth(const Point2d& pt) {
+    uint idx = FindNearestIdx(GetCenterLine(), pt);
+    // assumption: center, left and right have same # elements
+    Point2d left_pt = GetPointAtIdx(GetLeftBoundary(), idx);
```
patrickhart, line 59: this assumes that these points are always parallel.
A cleaner way would be to calculate the s-value and then query the point.
There is an interpolation available.

bark/world/world.cpp
```diff
@@ -244,6 +244,11 @@ FrontRearAgents World::GetAgentFrontRearForId(
     }
 
     FrenetPosition frenet_other(it->second->GetCurrentPosition(), center_line);
+    float width = lane_corridor->GetLaneWidth(it->second->GetCurrentPosition());
+    if (abs(frenet_other.lat) > width / 2) {
```
patrickhart, line 248: use std::abs
see: https://stackoverflow.com/questions/33738509/whats-the-difference-between-abs-and-fabs

bark/world/map/lane_corridor.hpp
klemense1: you mean calculating the s-value of the centerline? and then using this for querying the points for left and right boundary?
I would argue that the s-values of those boundaries are not necessarily corresponding to the s-val of the center line either. Thus, I am not sure it will improve the accuracy of the function, but will certainly make it slower...

PR #410, opened 2020-08-18
fix bug in base_idm if lane corridors are not same

klemense1:
- solves the issue that when of the distance calculation in idm_base `behavior.CalcNetDistance(agent1, agent2);` +
when agent2 has a different lane corridor than agent 1 (for example because it is on a successor road)
- previously, the frenet coordinates in this distance calculation where not comparable
- strange behavior occured for example in the merging scenario to the followint agent, if the leader was placed on the road after the merging point
- I added a unit test to show the bug

### DIFF
--- a/bark/models/behavior/idm/base_idm.cpp
+++ b/bark/models/behavior/idm/base_idm.cpp
@@ -106,8 +106,9 @@ double BaseIDM::CalcInteractionTerm(double net_distance, double vel_ego,
 }
 
 double BaseIDM::CalcNetDistance(
-    const std::shared_ptr<const Agent>& ego_agent,
+    const world::ObservedWorld& observed_world,
     const std::shared_ptr<const Agent>& leading_agent) const {
+  const auto& ego_agent = observed_world.GetEgoAgent();
   // relative velocity and longitudinal distance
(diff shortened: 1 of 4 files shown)


bark/models/behavior/idm/base_idm.cpp
```diff
@@ -117,7 +117,14 @@ double BaseIDM::CalcNetDistance(
   const State leading_state = leading_agent->GetCurrentState();
   const float other_velocity = leading_state(StateDefinition::VEL_POSITION);
 
-  FrenetPosition frenet_leading = leading_agent->CurrentFrenetPosition();
+  // we need to use the lane corridor of the ego agent to be able to compare the
+  // frenet values
+  const auto& lane_corridor =
```
patrickhart, line 122: Another idea would be to pass the observed_world& to `CalcNetDistance(...)`, then you could use
`const auto& lane_corridor = observed_world.GetLaneCorridor()` and `const auto& ego_agent = observed_world.GetEgoAgent()`

patrickhart: The idea and implementation itself makes sense!

PR #411, opened 2020-08-19
Fix lane width calc

klemense1:


bark/models/behavior/rule_based/mobil_behavior.cpp
```diff
@@ -86,10 +86,10 @@ BehaviorMobilRuleBased::ChooseLaneCorridor(
           GetVelocity(observed_world.GetEgoAgent()));
     }
   } else {
-    LOG(INFO) << "Target Corridor not available to MOBIL";
-    // Vehicle cannot stay in current corridor anymore, we thus incentivize a
-    // lane change
-    acc_ego = BaseIDM::GetAccelerationLowerBound();
+    // Target Corridor has been filtered out before as a corridor to change to
+    acc_ego = CalcLongRawAccWithoutLeader(
```
patrickhart, line 90: why use without leader here?
This assumes that we stay within the same lane corridor right? would the normal distance function of the IDM not be the best?

PR #412, opened 2020-08-20
Evaluator capture states

patrickhart:
Added an evaluator that captures all agent's states.

This enables to obtain all agents states by calling `world.Evaluate()` and allows, therefore, for state tracing.


klemense1 (approved):
seems fine for me, you may rename the evaluator to something like EvaluatorCaptureAgentStates, but that's minor.

PR #414, opened 2020-08-22
lat. filtering (FrontAgent) settable via parameter

klemense1:
As suggested by @juloberno  lateral filtering for GetFrontAndRearAgent based on lateral offset to center line can now be set via the parameter server. 
For frac_lateral_offset_ >= 1, it should work as before.

### DIFF
--- a/bark/world/world.cpp
+++ b/bark/world/world.cpp
@@ -27,7 +27,10 @@ World::World(const commons::ParamsPtr& params)
       remove_agents_(params->GetBool(
           "World::remove_agents_out_of_map",
           "Whether agents should be removed outside the bounding box.",
-          false)) {
+          false)),
+      frac_lateral_offset_(params->GetReal("World::FracLateralOffset",
+          "Fraction of lateral offset for FrontRearAgent Calculation, should be larger than 0.",
+          0.5)) {
   //! segfault handler
   std::signal(SIGSEGV, bark::commons::SegfaultHandler);
 }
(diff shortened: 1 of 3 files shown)


bark/world/world.hpp
```diff
@@ -128,6 +128,8 @@ class World : public commons::BaseType {
 
   bool GetRemoveAgents() { return remove_agents_; }
 
+  float GetFracLateralOffset() const { return frac_lateral_offset_; }
```
patrickhart, line 131: in general should be not use double as a few functions below we use that. I do not like the mixing of types

patrickhart (approved):
A general remark, should this not include the whole lane and if you are inbetween two lanes (say on the dotted line) you would want the minimum distance of both lanes?
In terms of autonomous driving this would make the most sense to me.

bark/world/world.cpp
```diff
@@ -27,7 +27,10 @@ World::World(const commons::ParamsPtr& params)
       remove_agents_(params->GetBool(
           "World::remove_agents_out_of_map",
           "Whether agents should be removed outside the bounding box.",
-          false)) {
+          false)),
+      frac_lateral_offset_(params->GetReal("World::FracLateralOffset",
```
juloberno, line 31: Gäbe es eine Möglichkeit, dass je nach BehaviorModel unterschiedlich parametrierbar zu machen?

Dadurch könnte man "nettere" und "weniger" nette Fahrer einfach über den Parameter definieren.

Ansonsten passt es aber für mich so.

PR #424, opened 2020-09-22
Update road_corridor_test.xodr to have left lane

ThrashAbaddon:
My colleague found out that file `road_corridor_test.xodr` is missing left lane linkage from road:0 lane:1 to road:1 lane:1. When using it with carla it causes a segfault.

This PR is adding missing linkage info.


patrickhart:
Will be merged in PR #425. Thanks for the contribution!

PR #419, opened 2020-08-27
Rss integration

tin1254:
# Pull overview

* Build RSS and the dependencies by bazel
    * build rss related files as an option by `--define rss=true`
* Implement rss check as an evaluator
    * implement a rss_interface to access rss libraries
* Implement visualization for rss safety response
* Implement drawing rss debugging message in viewer.py
* Add two examples running rss as a demo
* Implement python and C++ unit tests for rss
* Documentation

Rss can be now run on:
* city_highway_straight
* DR_DEU_Merging_MT_v01
* DR_CHN_Merging_ZS_partial_v02

PR #426, opened 2020-09-25
Rss integration (cherry-picked)

tin1254:
This RP contains rss integration commits extracted from #419 
This new PR is created because somehow the commits in that branch are messed up

# Pull overview

* Build RSS and the dependencies by bazel
    * build rss related files as an option by `--define rss=true`
* Implement rss check as an evaluator
    * implement a rss_interface to access rss libraries
* Implement visualization for rss safety response
* Implement drawing rss debugging message in viewer.py
* Add two examples running rss as a demo
* Implement python and C++ unit tests for rss
* Documentation

Rss can be now run on:
* city_highway_straight
* DR_DEU_Merging_MT_v01
* DR_CHN_Merging_ZS_partial_v02


tin1254:
The city highway map is centered, you could re-check the related code with the new map @klemense1 @patrickhart @juloberno

PR #437, opened 2020-10-20
Implementation of BehaviorSafety and BehaviorRSSConformant models

patrickhart:
**BehaviorSafety**
A model that is to be executed when e.g. the RSS has been violated.
It steers to the initial lane and brakes to a lower speed (almost zero).

**BehaviorRSSConformant**
Holds a nominal and safety behavior as well as an RSS evaluator. 
In every time-step it is being checked whether the RSS has been violated. If yes, then the BehaviorSafety model is executed.
Additionally, the time of the violation is stored as member.
It also sets the initial LaneCorridor of the BehaviorSafety model

**Testing**
We assume the RSS evaluator to work properly and only perform a system level check.
Two times the vehicle performs a lane change, but one time the RSS is (fake) violated and the lane-change maneuver is aborted.


tobiaskessler:
Assumption: The safety lane (corridor) is always the one that was active in the first simulation step -> ok atm, but not generic. In the next step we could reset the initial_lane_corr_ to the current lane corridor after a successful lane merge...

bark/models/behavior/rss_behavior/rss_behavior.hpp
```diff
@@ -0,0 +1,91 @@
...
+
+  void SetEvaluator(const std::shared_ptr<BaseEvaluator>& evaluator){
+    rss_evaluator_ = evaluator;
+  }
+
+ private:
+  std::shared_ptr<BehaviorModel> nominal_behavior_model_;
+  std::shared_ptr<BehaviorSafety> safety_behavior_model_;
+  // TODO: needs to be the RSS evaluator
+  std::shared_ptr<BaseEvaluator> rss_evaluator_;
```
klemense1, line 75: can be changed now :-)

klemense1 (approved):
looks good, see my comments

bark/models/behavior/rss_behavior/rss_behavior.cpp
```diff
@@ -0,0 +1,69 @@
...
+    LOG(INFO) << "Agent " << observed_world.GetEgoAgentId()
+              << ": Behavior status has expired!" << std::endl;
+    SetBehaviorStatus(BehaviorStatus::EXPIRED);
+    return GetLastTrajectory();
+  }
+
+  // TODO: cast and assert that it is the RSS evaluator
+  auto eval_res = boost::get<bool>(rss_evaluator_->Evaluate(observed_world));
+  if (eval_res) {
+    LOG(INFO) << "RSS is violated." << std::endl;
```
juloberno, line 38: Better use VLOG with some verbosity level above 2 here.

bark/models/behavior/rss_behavior/rss_behavior.hpp
juloberno: ensure here that the rss evalautor gets appropriate params passed with param server or you pass the evaluator in the constructor?

bark/models/behavior/safety_behavior/safety_behavior.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+#include "bark/models/behavior/safety_behavior/safety_behavior.hpp"
+#include "bark/models/behavior/idm/idm_lane_tracking.hpp"
+#include "bark/world/observed_world.hpp"
+
+namespace bark {
+namespace models {
+namespace behavior {
+
+Trajectory BehaviorSafety::Plan(
```
juloberno, line 20: same name as file names

bark/models/behavior/safety_behavior/safety_behavior.hpp
```diff
@@ -0,0 +1,78 @@
...
+  void SetInitialLaneCorridor(const LaneCorridorPtr& lc){
+    initial_lane_corr_ = lc;
+  }
+
+  ParamsPtr GetBehaviorSafetyParams() const {
+    return safety_behavior_params_;
+  }
+
+ private:
+  std::shared_ptr<BehaviorModel> behavior_model_;
```
juloberno, line 63: better naming required for this member. Change to specific IDM type or derive from this type.

bark/models/behavior/safety_behavior/safety_behavior.cpp
```diff
@@ -0,0 +1,46 @@
...
+    float min_planning_time, const world::ObservedWorld& observed_world) {
+  SetBehaviorStatus(BehaviorStatus::VALID);
+
+  // for now we support only the BehaviorIDMLaneTracking
+  BARK_EXPECT_TRUE(std::dynamic_pointer_cast<BehaviorIDMLaneTracking>(behavior_model_));
+
+  std::shared_ptr<BehaviorIDMLaneTracking> idm_lane_tracking_behavior =
```
juloberno, line 27: If you do a cast here without invalidity check, you should change the member pointer directly to the required type. Why do we need a pointer at all? Why not just derive from BehaviorIDMLaneTracking? In the constructor we already overwrite specific parameters for BehaviorIDMLaneTracking.

bark/models/behavior/safety_behavior/safety_behavior.hpp
```diff
@@ -0,0 +1,78 @@
...
+  LaneCorridorPtr initial_lane_corr_;
+  ParamsPtr safety_behavior_params_;
+};
+
+inline std::shared_ptr<BehaviorModel> BehaviorSafety::Clone() const {
+  std::shared_ptr<BehaviorSafety> model_ptr =
```
juloberno, line 69: clone function does not clone underlying behavior model, maybe this should be the case?

bark/models/tests/behavior_rss_test.cc
```diff
@@ -0,0 +1,172 @@
...
+  // safety behavior
+  auto params = std::make_shared<SetterParams>();
+  auto behavior_lane_tracking = std::make_shared<BehaviorIDMLaneTracking>(
+    params);
+  std::shared_ptr<BehaviorSafety> safety_behavior =
+    std::make_shared<BehaviorSafety>(params);
+  safety_behavior->SetBehaviorModel(behavior_lane_tracking);
+
+  // rss behavior
+  auto rss_behavior = BehaviorRSSConformant(params);
```
juloberno, line 75: Interface is not straightforward since object is not functioning by just calling the constructor. Either default construction of underlyinng behaviors or extend the constructor

juloberno (approved):

bark/python_wrapper/models/behavior.cpp
```diff
@@ -387,7 +389,50 @@ void python_behavior(py::module m) {
...
+  py::class_<BehaviorSafety, BehaviorModel,
+             shared_ptr<BehaviorSafety>>(m, "BehaviorSafety")
+    .def(py::init<const bark::commons::ParamsPtr&>())
+    .def("SetBehaviorModel", &BehaviorSafety::SetBehaviorModel)
+    .def("__repr__",
+      [](const BehaviorSafety& b) {
+        return "bark.behavior.BehaviorSafety";
+      })
+    .def(py::pickle(
+      [](const BehaviorSafety& b) {
```
juloberno, line 402: Does not serialize underlying behavior model (e.g. IDMLaneTracking) -> Segfault in Benchmarking...

bark/python_wrapper/models/behavior.cpp
```diff
@@ -387,7 +389,50 @@ void python_behavior(py::module m) {
...
+    .def("SetNominalBehaviorModel", &BehaviorRSSConformant::SetNominalBehaviorModel)
+    .def("SetSafetyBehaviorModel", &BehaviorRSSConformant::SetSafetyBehaviorModel)
+    .def("__repr__",
+      [](const BehaviorRSSConformant& b) {
+        return "bark.behavior.BehaviorRSSConformant";
+      })
+    .def(py::pickle(
+      [](const BehaviorRSSConformant& b) {
+        return py::make_tuple(
+          ParamsToPython(b.GetParams()));
```
juloberno, line 426: Does not serialize safety and nominal behavior models -> segfaults in benchmarking

juloberno (changes requested):

bark/python_wrapper/models/behavior.cpp
patrickhart: But the models will be set in the constructor to the default values?

bark/python_wrapper/models/behavior.cpp
patrickhart: But the models will be set in the constructor to the default values?

bark/models/tests/behavior_rss_test.cc
patrickhart: yes, it does?

bark/models/behavior/safety_behavior/safety_behavior.hpp
patrickhart: the default constructor will be called atm, which I think is okay?

bark/models/behavior/safety_behavior/safety_behavior.cpp
patrickhart: well, in case we want to use another one in the future? e.g. MCTS safety behavior or motion primitives or..

bark/models/behavior/safety_behavior/safety_behavior.hpp
patrickhart: would leave it generic see comment below

bark/models/behavior/safety_behavior/safety_behavior.cpp
patrickhart: The RSS evaluator implementation needs some rework.. I do not know what agent_id there is and also cannot handle observed world @klemense1

bark/models/behavior/rss_behavior/rss_behavior.hpp
patrickhart: done.

bark/models/behavior/rss_behavior/BUILD
```diff
@@ -0,0 +1,23 @@
+cc_library(
+    name = "rss_behavior",
+    srcs = [
+        "rss_behavior.cpp",
```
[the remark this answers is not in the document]
patrickhart: do not get this comment

bark/models/behavior/idm/idm_lane_tracking.cpp
```diff
@@ -47,7 +47,13 @@ std::tuple<Trajectory, Action> BehaviorIDMLaneTracking::GenerateTrajectory(
   dynamic::State ego_vehicle_state = observed_world.CurrentEgoState();
   double start_time = observed_world.GetWorldTime();
   double t_i = 0., acc = 0.;
-  geometry::Line line = lane_corr->GetCenterLine();
+  geometry::Line line;
+  if (constant_lane_corr_ != nullptr) {
```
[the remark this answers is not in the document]
patrickhart: well it is just to set a constant LaneCorridor. The naming might not be great, sure documentation should be added; also for the EvaluatorRSS ; )

bark/models/behavior/rss_behavior/rss_behavior.hpp
tobiaskessler: rather in the constructor?

bark/python_wrapper/models/behavior.cpp
tobiaskessler: ? how to resolve?

bark/models/behavior/safety_behavior/safety_behavior.cpp
juloberno: Okay, but how you implemented the safety behavior at the moment, it always requires the function SetConstantLaneCorridor(), when we use an MCTS we will rather replace the safety maneuver with the MCTS in the RSSConformantBehaviorModel instead of using an MCTS here.

bark/models/behavior/safety_behavior/safety_behavior.hpp
juloberno: yes, but it does not clone the underlying behavior model which (may) cause difficult to find bugs.

bark/models/behavior/rss_behavior/rss_behavior.hpp
tobiaskessler: does the rss evaluator has parameters at all?

bark/models/tests/behavior_rss_test.cc
juloberno: But, you always have to call the setters for it to work or not?

bark/python_wrapper/models/behavior.cpp
juloberno: Yes, but we need serialization of them. Otherwise we can only benchmark the defaults.

bark/models/behavior/safety_behavior/safety_behavior.cpp
tobiaskessler: Isnt this discussion a bit academic? :)

tobiaskessler (approved):
Works for me

bark/python_wrapper/models/behavior.cpp
juloberno: Implement getters for nominal and safety models and use them in the serialization:
py::make_tuple(
          ParamsToPython(b.GetParams()), ParamsToPython(b.GetNominalModel()->GetParams(), ParamsToPython(b.GetSafetyModel()->GetParams());

For deserialization, build recreate the nomal and safety models and use the setters to set them for RSSConformantModel

bark/models/behavior/safety_behavior/safety_behavior.cpp
juloberno: Actually it makes things more complicated for serialization than they have to be a.t.m. Thus, for me its not academic but introducing more work :-)

bark/models/behavior/safety_behavior/safety_behavior.cpp
klemense1: I just implemented and pushed the `Evaluate(ObservedWorld observed_worrld)` interface

bark/models/tests/behavior_rss_test.cc
patrickhart: No, they are also set by default. However, you can overwrite them using the setters

bark/models/behavior/rss_behavior/rss_behavior.hpp
patrickhart: as we do use the observed world, not anymore.

klemense1: we will eventually make the parameters settable via a parameter server. but not until tomorrow...

bark/models/behavior/safety_behavior/safety_behavior.cpp
tobiaskessler: For now, I de/serialized the pointer individually, but I see your point.

bark/models/behavior/safety_behavior/safety_behavior.hpp
tobiaskessler: I added a copy constructor

juloberno (approved):

PR #442, opened 2020-10-21
 Fix matplot deprecation warning

BastianHofmann:
Fix matplot deprecation warning in rss merging example.
```
MatplotlibDeprecationWarning: Case-insensitive properties were deprecated in 3.3 and support will be removed two minor releases later
```


patrickhart:
PR #444

PR #454, opened 2020-11-09
RSS Check if observed world is found before accessing

BastianHofmann:
This occurs with the interaction dataset sometimes.

### DIFF
--- a/bark/world/evaluation/rss/evaluator_rss.hpp
+++ b/bark/world/evaluation/rss/evaluator_rss.hpp
@@ -75,8 +75,11 @@ class EvaluatorRSS : public BaseEvaluator {
   // lateral one is unsafety.
   virtual EvaluationReturn Evaluate(const World& world) {
     WorldPtr cloned_world = world.Clone();
-    ObservedWorld observed_world = cloned_world->Observe({agent_id_})[0];
-    return rss_.GetSafetyReponse(observed_world);
+    std::vector<ObservedWorld> observed_worlds = cloned_world->Observe({agent_id_});
+    if (observed_worlds.size() > 0) {
+      return rss_.GetSafetyReponse(observed_worlds[0]);
+    }
+    return false;
   };
 
   virtual EvaluationReturn Evaluate(const ObservedWorld& observed_world) {


bark/world/evaluation/rss/evaluator_rss.hpp
```diff
@@ -75,8 +75,11 @@ class EvaluatorRSS : public BaseEvaluator {
   // lateral one is unsafety.
   virtual EvaluationReturn Evaluate(const World& world) {
     WorldPtr cloned_world = world.Clone();
-    ObservedWorld observed_world = cloned_world->Observe({agent_id_})[0];
-    return rss_.GetSafetyReponse(observed_world);
+    std::vector<ObservedWorld> observed_worlds = cloned_world->Observe({agent_id_});
+    if (observed_worlds.size() > 0) {
+      return rss_.GetSafetyReponse(observed_worlds[0]);
```
juloberno, line 80: The problem is that an evaluation is requested for an agent already out of the map or not yet valid. This should be caught already one lever higher not defining an evaluator with agent_id_ when agent_id_ is not in world.GetAgents().

If nevertheless someone calls evaluate with an unvailable agent_id_ you should add a warning message with LOG(WARNING)

juloberno (changes requested):

bark/world/evaluation/rss/evaluator_rss.hpp
BastianHofmann: @juloberno I checked again and the observed_worlds will be empty if the only agent in the scene is the ego agent (see https://github.com/bark-simulator/bark/blob/master/bark/world/world.cpp#L141). If I understand this line correctly, `agents_.end()` will be the ego agent and we will return an empty `observed_worlds` vector.

BastianHofmann: @juloberno I can still add the warning if you want to, but to me it seems like this is expected behaviour. If the only agent in the scene is the ego agent, RSS cannot be violated (i.e. we return false from the `Evaluate` method).

PR #457, opened 2020-11-13
modify runtime.step function to allow for correct plotting

klemense1:
To allow for a correct plotting of planning information, I suggest to split the world.step() function within runtime.step().
This way, we generate the plans (via PlanAgents()) and can then plot it within the viewer.world() function. 
Only then, all agents shall move (via Execute). Otherwise, we are always plotting a planning result that does not match the world figure.

### DIFF
--- a/bark/runtime/runtime.py
+++ b/bark/runtime/runtime.py
@@ -37,10 +37,16 @@ class Runtime(PyRuntime):
     self._viewer.Reset()
 
   def step(self):
-    assert(self._reset_has_been_called==True)
-    self._world.Step(self._step_time)
+    assert(self._reset_has_been_called == True)
+    # instead of calling world.step() seperately, we call PlanAgents() and Execute() seperately
+    self._world.PlanAgents(self._step_time)
+
     if self._render:
-      self.render()
+        self.render()
+
+    inc_world_time = self._world.time + self._step_time
+    self._world.Execute(inc_world_time)
+    self._world.time = inc_world_time
 
   def render(self):
     # self._viewer.clear()


bark/runtime/runtime.py
```diff
@@ -37,10 +37,16 @@ def reset(self, scenario=None):
...
   def step(self):
-    assert(self._reset_has_been_called==True)
-    self._world.Step(self._step_time)
+    assert(self._reset_has_been_called == True)
+    # instead of calling world.step() seperately, we call PlanAgents() and Execute() seperately
+    self._world.PlanAgents(self._step_time)
+
     if self._render:
-      self.render()
+        self.render()
```
patrickhart, line 45: the question is what should be plotted. The execute function is what the agent's state is actually set to in the next time-step.

So by plotting only the planned state there could be errors and deviations.
My suggestion would be to add a feature in the viewer called `VisualizePlannedTrajectory()` or `VisualizePlannedTrajectory`.

klemense1: I would not suggest to plot the vehicles at the planned states. But allow to plot VisualizePlannedTrajectory() as you just said. But if we do not call this after world.step, the states are updated, so the planning results are basically outdated.

klemense1: basically, what we are doing currently is plotting the world too late anyway. We never plot the world at t=0, as we are rendering/ plotting always AFTER we step the world.

patrickhart: for t=0, we should plot the world before the runtime cycle.
so in the end of the reset function (scenario generation), we should call the render function.
I would leave the functionality as is, but add the option to draw the planned state and trajectory.

PR #460, opened 2020-11-17
Steering Angle and FrenetState Fix

klemense1:
- defined angles in [-pi, -pi]
- fix FrenetState calculation
- reverting unwanted changes in runtime

### DIFF
--- a/bark/models/behavior/rule_based/intersection_behavior.cpp
+++ b/bark/models/behavior/rule_based/intersection_behavior.cpp
@@ -28,8 +28,8 @@ namespace behavior {
 
 using bark::commons::SetterParams;
 using bark::commons::transformation::FrenetPosition;
-using bark::geometry::Norm0To2PI;
 using bark::geometry::Point2d;
+using bark::geometry::AngleDiff;
 using bark::geometry::SignedAngleDiff;
 using bark::models::behavior::BehaviorConstantAcceleration;
 using bark::models::dynamic::DynamicModelPtr;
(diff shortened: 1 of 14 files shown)


bark/models/behavior/rule_based/intersection_behavior.cpp
```diff
@@ -76,9 +76,9 @@ std::pair<AgentId, bool> BehaviorIntersectionRuleBased::GetIntersectingAgent(
           observed_world.GetLaneCorridor()->GetCenterLine(), ego_pos));
       double s_other = std::get<1>(GetNearestPointAndS(
           observed_world.GetLaneCorridor()->GetCenterLine(), agent_pos));
-      double ego_angle = Norm0To2PI(ego_state[THETA_POSITION]);
-      double other_angle = Norm0To2PI(agent_state[THETA_POSITION]);
-      if (fabs(ego_angle - other_angle) > angle_diff_for_intersection_ &&
+      double ego_angle = ego_state[THETA_POSITION];
+      double other_angle = agent_state[THETA_POSITION];
+      if (AngleDiff(ego_angle, other_angle) > angle_diff_for_intersection_ &&
```
patrickhart, line 81: I think a signed angle diff is required here actually.
if both angles are in the range of [-pi, pi], the difference could be -0.2 or 0.2 which both would equal to 0.2.
So we loose the information of the direction which is required as we only want to yield vehicles from the right

patrickhart (approved):

bark/models/behavior/rule_based/intersection_behavior.cpp
klemense1: you are right

PR #461, opened 2020-11-17
Fix RSS evaluator if agent is alone in scene

BastianHofmann:
The original PR has been closed, because the target branch was merged.

This occurs with the interaction dataset sometimes.

Refer to this comment to why I think this is expected behaviour and should not have a warning: https://github.com/bark-simulator/bark/pull/454#discussion_r525100246


klemense1:
merged already in other PR

PR #462, opened 2020-11-17
Flexible RSS parameters and all double in BARK

patrickhart:
- The ParameterServer() is now passed to the RSSInterface directly making it possible to specify parameters more flexible.
- Everything is now implemented using double


patrickhart:
Sorry for the long PR, but it seems not all floats had been replaced by doubles

bark/models/tests/behavior_idm_classic_test.cc
```diff
@@ -252,32 +252,32 @@ TEST(coolness_factor_upper_eq_case, behavior_idm_classic) {
...
   // First case, we start with the desired velocity. After num steps, we should
   // advance
-  float ego_velocity = desired_velocity, rel_distance = 10.0,
-        velocity_difference = -2, acc_ego = 2.0f, acc_other = -5.0f;
-  float other_velocity = ego_velocity - velocity_difference;
-  float time_step = 0.2f;  // Very small time steps to verify differential
+  double ego_velocity = desired_velocity, rel_distance = 10.0,
+        velocity_difference = -2, acc_ego = 2.0, acc_other = -5.0;
+  double other_velocity = ego_velocity - velocity_difference;
+  double time_step = 0.2f;  // Very small time steps to verify differential
```
klemense1, line 280: you could also replace this by 0.2

bark/models/tests/behavior_idm_classic_test.cc
```diff
@@ -316,32 +316,32 @@ TEST(coolness_factor_lower_eq_case_vel_diff_neg, behavior_idm_classic) {
...
   // First case, we start with the desired velocity. After num steps, we should
   // advance
-  float ego_velocity = desired_velocity, rel_distance = 20.0,
-        velocity_difference = -1, acc_ego = 2.0f, acc_other = 1.0f;
-  float other_velocity = ego_velocity - velocity_difference;
-  float time_step = 0.2f;  // Very small time steps to verify differential
+  double ego_velocity = desired_velocity, rel_distance = 20.0,
+        velocity_difference = -1, acc_ego = 2.0, acc_other = 1.0;
+  double other_velocity = ego_velocity - velocity_difference;
+  double time_step = 0.2f;  // Very small time steps to verify differential
```
klemense1, line 344: same here

bark/models/tests/behavior_idm_classic_test.cc
```diff
@@ -375,28 +375,28 @@ TEST(coolness_factor_lower_eq_case_vel_diff_pos, behavior_idm_classic) {
...
   params->SetReal("BehaviorIDMClassic::AccelerationUpperBound", 1000.0);
-  params->SetReal("BehaviorIDMClassic::DesiredVelocity", 8.0f);
-  params->SetReal("BehaviorIDMClassic::ComfortableBrakingAcceleration", 1.0f);
-  params->SetReal("BehaviorIDMClassic::MinVelocity", 0.0f);
-  params->SetReal("BehaviorIDMClassic::MaxVelocity", 50.0f);
+  params->SetReal("BehaviorIDMClassic::DesiredVelocity", 8.0);
+  params->SetReal("BehaviorIDMClassic::ComfortableBrakingAcceleration", 1.0);
+  params->SetReal("BehaviorIDMClassic::MinVelocity", 0.0);
+  params->SetReal("BehaviorIDMClassic::MaxVelocity", 50.0);
   params->SetReal("BehaviorIDMClassic::CoolnessFactor", 0.6f);
```
klemense1, line 388: same here

klemense1 (approved):

bark/world/evaluation/rss/rss_interface.hpp
```diff
@@ -90,32 +90,45 @@ typedef std::unordered_map<objects::AgentId, std::pair<bool, bool>>
...
+    acc_lon_max_ = params->GetReal(
+      "EvaluatorRss::AccLonMax", "maximum acceleration", 1.7);
+    brake_lon_max_ = params->GetReal(
+      "EvaluatorRss::BrakeLonMax", "maximum deceleration", -1.7);
+    brake_lon_min_ = params->GetReal(
+      "EvaluatorRss::BrakeLonMin", "minimum braking deceleration", -1.69);
+    brake_lon_min_correct_ = params->GetReal(
+      "EvaluatorRss::BrakeLonMinCorrect",
+      "minimum deceleration of oncoming vehicle", -1.67);
+    acc_lat_brake_max_ = params->GetReal(
```
juloberno, line 104: I would prefer that we have to completely different sets of parameters for ego and others. Not only the response time. Since there may be request in the future for instance, to have different acc. limits for ego and others.

juloberno (commented):
Looks good. In general, I was always annoyed by this double vs. float. Now, with the change we should have directly introduced a typedef then replacing for all floating point data types.

bark/world/evaluation/rss/rss_interface.hpp
patrickhart: Was thinking about that as well!
My suggestion would be to introduce a struct class and then have two of these as members containing all parameters

PR #463, opened 2020-11-18
Acc corridor rss

klemense1:
- implemented struct AccelerationLimits in BaseIDM
- is usually constructed in constructor once but can be set during runtime (e.g. by BehaviorRSS)
- acceleration corridor is read from RssResponse and parsed to AccelerationLimits


bark/models/behavior/idm/base_idm.cpp
```diff
@@ -39,12 +39,12 @@ BaseIDM::BaseIDM(const commons::ParamsPtr& params) : BehaviorModel(params) {
                       "See Wikipedia IDM article", 1.5f);
   param_max_acceleration_ = params->GetReal(
       "BehaviorIDMClassic::MaxAcceleration", "See Wikipedia IDM article", 1.7f);
-  param_acceleration_lower_bound_ =
-      params->GetReal("BehaviorIDMClassic::AccelerationLowerBound",
-                      "See Wikipedia IDM article", -5.0f);
-  param_acceleration_upper_bound_ =
-      params->GetReal("BehaviorIDMClassic::AccelerationUpperBound",
-                      "See Wikipedia IDM article", 8.0f);
+  // param_acceleration_lower_bound_ =
```
patrickhart, line 42: should we not maintain both options:
- one where we restrict the accelerations as we desire
- and one where we strictly adhere to the RSS accelerations

I think especially for the other vehicles this could be interesting if not everyone behaves strictly according to the rss parameters (mixed-traffic).

bark/models/behavior/idm/base_idm.hpp
```diff
@@ -97,14 +99,14 @@ class BaseIDM : virtual public BehaviorModel {
   const float GetDesiredTimeHeadway() const {
     return param_desired_time_head_way_;
   }  // unit is seconds
-  const float GetMaxAcceleration() const {
+  const float GetLonAccelerationMax() const {
     return param_max_acceleration_;
```
patrickhart, line 103: should we not rename the member in `param_max_lon_acceleration_` then?

bark/models/behavior/idm/base_idm.hpp
```diff
@@ -97,14 +99,14 @@ class BaseIDM : virtual public BehaviorModel {
...
   }  // unit is meter/second^2
   const float GetAccelerationLowerBound() const {
-    return param_acceleration_lower_bound_;
+    return acceleration_limits_.lon_acc_min;
```
patrickhart, line 106: same here

bark/models/behavior/idm/base_idm.hpp
```diff
@@ -97,14 +99,14 @@ class BaseIDM : virtual public BehaviorModel {
...
   }
   const float GetAccelerationUpperBound() const {
-    return param_acceleration_upper_bound_;
+    return acceleration_limits_.lon_acc_max;
```
patrickhart, line 109: and here in `GetAccelerationLonLowerBound()` and `GetAccelerationLonUpperBound()`

bark/models/dynamic/single_track.hpp
```diff
@@ -106,13 +125,19 @@ inline double CalculateSteeringAngle(const SingleTrackModelPtr& model,
...
+    double delta_max_right =
+        std::min(model->GetSteeringAngleMax(),
+                 std::abs(std::atan2(model->GetLatAccelerationRightMax() *
+                                         model->GetWheelBase(),
+                                     vel * vel)));
+    double delta_max_left =
         std::min(model->GetSteeringAngleMax(),
                  std::abs(std::atan2(
-                     model->GetLatAccelerationMax() * model->GetWheelBase(),
+                     model->GetLatAccelerationLeftMax() * model->GetWheelBase(),
```
patrickhart, line 137: Why Left? -> Side?

bark/models/tests/dynamic_test.cc
```diff
@@ -84,26 +84,28 @@ TEST(valid_trajectory_test, dynamic_test) {
   EXPECT_FALSE(IsValid(traj4));
 }
 
+double CalcAccLat(const State& x, const State& x1, const double dt) {
+  auto theta_dot = (x1(static_cast<int>(StateDefinition::THETA_POSITION)) -
+                    x(static_cast<int>(StateDefinition::THETA_POSITION))) /
+                    dt;
+  return theta_dot * x(static_cast<int>(StateDefinition::VEL_POSITION));
```
patrickhart, line 91: theta_dot [rad/s] * v [m/s]. Is that correct?

Could we not just: acceleration * cos(theta)?

patrickhart (approved):

bark/models/behavior/behavior_rss/behavior_rss.cpp
```diff
@@ -75,6 +77,7 @@ Trajectory BehaviorRSSConformant::Plan(
   dynamic::Trajectory last_traj;
   if (behavior_rss_status_ == BehaviorRSSConformantStatus::NOMINAL_BEHAVIOR) {
     // execute normal
+    ApplyRestrictionsToNominalModel();
```
juloberno, line 80: It is somewhat hidden that we use a member variable here to set the restrictions. Better pass the restrictions to the function.

bark/models/behavior/behavior_rss/behavior_rss.cpp
```diff
@@ -89,7 +92,22 @@ Trajectory BehaviorRSSConformant::Plan(
   return last_traj;
 }
 
-
+void BehaviorRSSConformant::ApplyRestrictionsToNominalModel() {
+  #ifdef RSS
```
juloberno, line 96: this ifdef should be not in this function -> better CALL this function only when rss is defined. Thus calling it in Plan in the ifdef would be better

bark/models/dynamic/single_track.hpp
klemense1: yes, that's what we get from the RSSResponse

juloberno (commented):
Looks good, however, for someone unfamiliar with the concept in the Plan function of BehaviorRSSConformant, the ifdefs and the hidden information passing via members may be confusing.

PR #470, opened 2020-11-23
Extend scenario dumper for the OpenScenario Trajectory defintion

patrickhart:
This implements a basic OpenScenario-Trajectory export functionality for the scenario dumper.


bark/benchmark/scenario_dumper.py
```diff
@@ -110,3 +111,56 @@ def write_map(self, config_idx, folder):
...
+      base_result_folder, benchmark_result)
+  
+  def GetTrajectoryPerAgent(self, world):
+    trajectory_per_agents = {}
+    for (agent_id, agent) in world.agents.items():
+      trajectory_per_agents[agent_id] = []
+      for state_action in agent.history:
+        state = state_action[0]
+        trajectory_per_agents[agent_id].append(
+          [state[0], state[1], state[2], state[3], state[4]])
```
klemense1, line 128: maybe use state definition instead of plain index?

bark/benchmark/scenario_dumper.py
```diff
@@ -110,3 +111,56 @@ def write_map(self, config_idx, folder):
...
+class OpenScenarioDumper(ScenarioDumper):
+  def __init__(self, base_result_folder, benchmark_result):
+    super(OpenScenarioDumper, self).__init__(
+      base_result_folder, benchmark_result)
+  
+  def GetTrajectoryPerAgent(self, world):
+    trajectory_per_agents = {}
+    for (agent_id, agent) in world.agents.items():
+      trajectory_per_agents[agent_id] = []
+      for state_action in agent.history:
```
klemense1, line 125: if history is too small, we are not saving everything. Do we have a mechanism to check for that?

bark/benchmark/scenario_dumper.py
```diff
@@ -110,3 +111,56 @@ def write_map(self, config_idx, folder):
...
+    world = scenario.GetWorldState()
+    trajectory_per_agents = self.GetTrajectoryPerAgent(world)
+    temp_vertex, temp_xml = self.GetTemplates()
+    
+    for agent_id, traj in trajectory_per_agents.items():
+      agent_xml = etree.fromstring(temp_xml)
+      polyline = agent_xml.findall(".//Polyline")[0]
+      for state in traj:
+        vertex = etree.fromstring(temp_vertex)
+        vertex.attrib['time'] = str(state[0])
```
klemense1, line 157: same here, use state definition

bark/benchmark/scenario_dumper.py
```diff
@@ -110,3 +111,56 @@ def write_map(self, config_idx, folder):
...
+    scenario = histories[-1]  # the last state inclues all the historic states
+    world = scenario.GetWorldState()
+    trajectory_per_agents = self.GetTrajectoryPerAgent(world)
+    temp_vertex, temp_xml = self.GetTemplates()
+    
+    for agent_id, traj in trajectory_per_agents.items():
+      agent_xml = etree.fromstring(temp_xml)
+      polyline = agent_xml.findall(".//Polyline")[0]
+      for state in traj:
+        vertex = etree.fromstring(temp_vertex)
```
klemense1, line 156: maybe put this in a seperate functkion StateToVertex?

klemense1 (changes requested):
Do we have a unit test for that?

bark/benchmark/scenario_dumper.py
patrickhart: I admit I just copied and pasted this from the original implementation. Will change that!

bark/benchmark/scenario_dumper.py
patrickhart: mhm, but where is defined what the expected history is? I also copy and pasted the cod here.

bark/benchmark/scenario_dumper.py
patrickhart: I thought of that, but as the function is still relatively short I did not think it is necessary

bark/benchmark/scenario_dumper.py
patrickhart: I have implemented a log message for now; but it also would be possible to throw an exception.

patrickhart:
> Do we have a unit test for that?

There was no unit test available for the ScenarioDumper in the first place. Even though I agree that it would be nice to have one, the export functionality rather represents an edge case and is not integral to BARK's functionality. Thus, this would be a question of prioritization to implement one.

klemense1 (approved):

PR #476, opened 2020-12-01
Rss acc and viewer

klemense1:
- Bugfix AppendLinesNoIntersect 
- Some Python Wrapping
- Extension for Runtime to extract time series of simulation
- Making shape in LaneCorridorConfig settable using radius and wheelbase
- Fixing LastAction for IDMLaneTracking
- Bugfix ValidAtTime Check in Execution 
- Acceleration Restrictions to Safety Model BehaviorRSSConformant

### DIFF
--- a/bark/examples/merging.py
+++ b/bark/examples/merging.py
@@ -100,7 +100,8 @@ sim_real_time_factor = param_server["simulation"]["real_time_factor",
 env = Runtime(step_time=0.2,
               viewer=viewer,
               scenario_generator=scenarios,
-              render=True)
+              render=True,
+              maintain_world_history=True)
 
 # run 3 scenarios
 for _ in range(0, 3):
@@ -109,5 +110,8 @@ for _ in range(0, 3):
   for step in range(0, 20):
     env.step()
     time.sleep(sim_step_time/sim_real_time_factor)
+    
+  df = env.ExtractTimeSeries()
+  print(df)
 
 # viewer.export_video(filename="/Users/hart/2019/bark/video/video", remove_image_dir=False)
--- a/bark/models/behavior/behavior_rss/behavior_rss.cpp
+++ b/bark/models/behavior/behavior_rss/behavior_rss.cpp
@@ -11,8 +11,8 @@
 #include <tuple>
 
 #include "bark/models/behavior/behavior_rss/behavior_rss.hpp"
-#include "bark/world/observed_world.hpp"
 #include "bark/models/dynamic/single_track.hpp"
+#include "bark/world/observed_world.hpp"
 
 namespace bark {
 namespace models {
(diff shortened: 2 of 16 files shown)


bark/models/behavior/idm/idm_lane_tracking.hpp
```diff
@@ -32,6 +32,9 @@ class BehaviorIDMLaneTracking : public BaseIDM {
     crosstrack_error_gain_ =
         params->GetReal("BehaviorIDMLaneTracking::CrosstrackErrorGain",
                         "Tuning factor of stanley controller", 1.0);
+    dynamic::Input input(2);
+    input << 0.0, 0.0;
+    SetLastAction(input);
```
patrickhart, line 37: why is this required in the constructor?

bark/world/world.cpp
```diff
@@ -72,7 +72,7 @@ void World::Execute(const double& delta_time) {
   const double inc_world_time = world_time_ + delta_time;
   using models::dynamic::StateDefinition::TIME_POSITION;
   for (auto agent : agents_) {
-    if (agent.second->IsValidAtTime(inc_world_time) &&
+    if (agent.second->IsValidAtTime(world_time_) &&
```
patrickhart, line 75: what does this change do? was it not working before?

patrickhart (approved):

bark/world/world.cpp
klemense1: that's only relevant when simulating the interaction dataset in closed loop. It's an error which I introduced when refactoring the Execute function. With this change now, it's the same as it was before.

bark/models/behavior/idm/idm_lane_tracking.hpp
klemense1: Cause this https://github.com/bark-simulator/bark/blob/master/bark/models/behavior/idm/base_idm.cpp#L78 needs to be overwritten after the  BaseIDM constructor is called in the BehaviorIDMLaneTracking constructor

bark/examples/merging.py
```diff
@@ -109,5 +110,8 @@ def goal(self, world):
   for step in range(0, 20):
     env.step()
     time.sleep(sim_step_time/sim_real_time_factor)
+    
+  df = env.ExtractTimeSeries()
+  print(df)
```
tobiaskessler, line 115: remove print?

bark/models/behavior/behavior_rss/behavior_rss.cpp
```diff
@@ -103,20 +114,36 @@ Trajectory BehaviorRSSConformant::Plan(
...
+
+void BehaviorRSSConformant::ApplyRestrictionsToSafetyModel(
+    const AccelerationLimits& limits) {
+  VLOG(4) << "AccelerationLimits for safety model " << limits;
+  std::shared_ptr<BehaviorIDMLaneTracking> safety_behavior_ptr =
+      std::dynamic_pointer_cast<BehaviorIDMLaneTracking>(
+          behavior_safety_model_);
+  safety_behavior_ptr->SetAccelerationLimits(limits);
+}
+
```
tobiaskessler, line 146: I think it is ok to have ApplyRestrictionsToNominalModel and ApplyRestrictionsToSafetyModel. But it could be one template function

tobiaskessler (approved):
looks good!

bark/examples/merging.py
klemense1: I guess for an example that's sufficient.

PR #477, opened 2020-12-01
SafetyPolygon for the RSS visualization and ScenarioDumper extension

patrickhart:
* Extension of the ScenarioDumper. It is now possible to specify config_ids
* Drawing of SafetyPolygons that contain the lateral and longitudinal safety distances, as well as other additional information
* Python-bindings for the SafetyPolygon
* Modification of the RSSViewer to plot the SafetyPolygons
* Modification of the RSSEvaluator to store additional data
* New interfaces for the RSSEvaluator, SafetyPolygon, and BehaviorRSSConformant
* Created and extended the serialization of SafetyPolygon and BehaviorRSSConformant

### DIFF
--- a/bark/world/evaluation/rss/safety_polygon.hpp
+++ b/bark/world/evaluation/rss/safety_polygon.hpp
@@ -1,0 +1,42 @@
+// Copyright (c) 2020 fortiss GmbH
+//
+// Authors: Julian Bernhard, Klemens Esterle, Patrick Hart and
+// Tobias Kessler
+//
+// This work is licensed under the terms of the MIT license.
+// For a copy, see <https://opensource.org/licenses/MIT>.
+
+#ifndef BARK_WORLD_EVALUATION_SAFETY_POLYGON_HPP_
+#define BARK_WORLD_EVALUATION_SAFETY_POLYGON_HPP_
+
+#include <limits>
+#include <memory>
+#include <string>
+
+#include "bark/geometry/polygon.hpp"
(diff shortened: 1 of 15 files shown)


juloberno (approved):
Looks good and can be merged. I especially like the idea to introduce the safety polygon as separate type.

bark/world/evaluation/rss/safety_polygon.hpp
```diff
@@ -0,0 +1,129 @@
...
+    return polygon;
+  }
+  AgentId GetAgentId() const {
+    return agent_id;
+  }
+};
+
+typedef std::shared_ptr<SafetyPolygon> SafetyPolygonPtr;
+
+inline void ComputeSafetyPolygon(
```
klemense1, line 50: what is this function doing? is it transforming a cartesian axis-aligned polygon to the orientation of the vehicle of the agent? can it be implemented using some base functions from bark::polygon (transform or so?)
either way, a doc-string might be beneficial.

klemense1 (changes requested):
Docstring for ComputeSafetyPolygon() would be good.

bark/world/evaluation/rss/safety_polygon.hpp
patrickhart: Okay, will add that. I checked and I did not find any appropriate geometry functions.
boost::geometry::scale_transformer e.g. transforms it in all dimensions.

klemense1 (approved):

PR #480, opened 2020-12-03
MinRssBrakingDistance for all Vehicles and Fixed ComputeSafetyPolygon

patrickhart:
* Fixes segfault issue with `ComputeSafetyPolygon` when the other agent has already been deleted
* Implements a calculation function for the minimum braking distance of an agent `ComputeMinBrakingPolygon`
* Modified viewer that draws all the braking distances using the flag `params["Visualization"]["Evaluation"]["DrawMinRSSBrakingDistances"] = True`

### DIFF
--- a/bark/runtime/viewer/viewer.py
+++ b/bark/runtime/viewer/viewer.py
@@ -352,7 +352,10 @@ class BaseViewer(Viewer):
               self.drawBehaviorPlan(eval_agent)
         
         if self._draw_ego_rss_safety_responses:
-          self.DrawRSSEvaluatorState(world, eval_agent_ids[0], eval_agent_ids[0])
+          self.DrawRSSEvaluatorState(world, eval_agent_ids[0])
+        
+        if self._rss_min_braking_distances:
+          self.DrawRSSBrakingDistances(world, eval_agent_ids[0])
 
     def drawMapAerialImage(self):
         pass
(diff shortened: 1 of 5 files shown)


bark/runtime/viewer/viewer.py
```diff
@@ -481,9 +489,26 @@ def drawRssSafetyResponses(self, world, ego_id):
             self.drawPolygon2d(transformed_polygon, response_color,
                                0.6, response_color, zorder=9)
 
+    def DrawRSSBrakingDistances(self, world, eval_id):
```
klemense1, line 492: does this also work when using the EvaluatorRSS and not the BehaviorRSSCompliant? Maybe add a docstring to indicate usage?

bark/world/evaluation/rss/safety_polygon.hpp
```diff
@@ -53,6 +53,18 @@ struct SafetyPolygon {
 
 typedef std::shared_ptr<SafetyPolygon> SafetyPolygonPtr;
 
+/**
+ * @brief  Modes for drawing the longitudinal visualization
+ * @note   
+ * @retval None
+ */
+enum LonDirectionMode {
+  AUTO = 0,
```
klemense1, line 62: what does auto mean?

bark/world/evaluation/rss/safety_polygon.hpp
```diff
@@ -102,24 +118,39 @@ inline void ComputeSafetyPolygon(
...
+    if (directional == LonDirectionMode::AUTO) {
+      // calculate if the agent is in front or behind
+      auto other_agent = observed_world.GetAgents()[safe_poly.agent_id];
+      if (!other_agent)
+        return;
+      auto other_pose = other_agent->GetCurrentPosition();
+      double relative_angle = atan2(
+        bg::get<1>(ego_pose) - bg::get<1>(other_pose),
+        bg::get<0>(ego_pose) - bg::get<0>(other_pose));
+      double diff_angle = SignedAngleDiff(theta - M_PI_2, relative_angle);
```
klemense1, line 131: why M_PI_2? is it then always within [-pi, pi]?

klemense1 (changes requested):
Nice! have a look on my comment towards the valid range of theta

bark/world/evaluation/rss/safety_polygon.hpp
patrickhart: this is a 90 degree turn

bark/world/evaluation/rss/safety_polygon.hpp
patrickhart: it either draws the polygon in front or behind based on the other agent. it is stated in the function itself

bark/runtime/viewer/viewer.py
patrickhart: Yes, it always works. Thus, it is nor required in my opinion

bark/world/evaluation/rss/commons.hpp
```diff
@@ -0,0 +1,110 @@
...
+  Acceleration brake_min_correct_lon = Acceleration(rss_params->GetReal(
+    "BrakeLonMinCorrect", "minimum deceleration of oncoming vehicle", -1.67));
+  Duration response_time = Duration(
+    rss_params->GetReal("TimeResponse", "response time of the ego vehicle", 0.2));
+  Distance min_stopping_distance_lat;
+  Acceleration accel_max_lat = Acceleration(
+    rss_params->GetReal("AccLatBrakeMax", "maximum lateral acceleration", 0.2));
+  Acceleration brake_min_correct_lat = Acceleration(
+    rss_params->GetReal("AccLatBrakeMin", "minimum lateral braking", -0.8));
+
```
tobiaskessler, line 69: are these the default values from Intel?

bark/world/evaluation/rss/safety_polygon.hpp
```diff
@@ -65,9 +77,13 @@ typedef std::shared_ptr<SafetyPolygon> SafetyPolygonPtr;
  * @retval None
  */
 inline void ComputeSafetyPolygon(
```
tobiaskessler, line 79: return bool if success or not?

tobiaskessler (approved):
looks good, only minor stuff, see my comments

bark/world/evaluation/rss/commons.hpp
patrickhart: Yes, these are the same parameters as in the BehaviorRSSConformant.
Ideally, these should be specified in a parameter file, but if not, these should still make sense.

klemense1 (approved):

PR #481, opened 2020-12-04
Change LaneCorridor selection in BaseIDM

patrickhart:
* if a static lane corridor is set, the LaneCorridor is now chosen using `ChooseLaneCorridorBasedOnVehicleState`
* ChooseLaneCorridorBasedOnVehicleState computes the lane corridor based on intersections and checks if any of the agent's polygon points in the front are in the other LaneCorridor

Observation: harsh braking does not occur directly at the beginning of the episode.

### DIFF
--- a/bark/world/map/commons.hpp
+++ b/bark/world/map/commons.hpp
@@ -60,17 +60,21 @@ inline LaneCorridorPtr ChooseLaneCorridorBasedOnVehicleState(
     }
 
     // if it is only in one lane corridor
-    if (other_lane_corr == nullptr)
-      return target_corr;
+    if (other_lane_corr == nullptr) {
+      // in this case we are entirely in one lane corridor
+      // here we want to only account for the preceeding vehicle
+      return observed_world.GetCurrentLaneCorridor();
+    }
 
-    // check if any point of the vehicle shape (front)
-    // is in the other lane corridor
+
+    // check if the front points of the ego vehicle are in the
+    // other lane corridor
     for (auto pt : vehicle_shape.obj_.outer()) {
       double pt_angle = atan2(
         bg::get<1>(ego_pose) - bg::get<1>(pt),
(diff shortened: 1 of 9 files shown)


bark/world/map/commons.hpp
```diff
@@ -0,0 +1,94 @@
...
+ * @brief  Function that selects LaneCorridors
+ * @note   If an agent is in two LaneCorridors with its shape
+ *         the other one it selected if the front points of the
+ *         agent are within the other lane corridor.
+ *         This models an entering of the other lane corridor.
+ * @param  observed_world: ObservedWorld of agent
+ * @param  target_corr: fallback corridor in case calculation fails
+ * @retval LaneCorridorPtr: the computed lane corridor
+ */
+inline LaneCorridorPtr ChooseLaneCorridorBasedOnVehicleState(
```
klemense1, line 42: I like it! Do we have a unit test for it?

klemense1 (changes requested):
Please consider writing a unit test for ChooseLaneCorridorBasedOnVehicleState. Also, did you check if BehaviorLaneChangeMobil is still working (in the examples)?

bark/world/map/commons.hpp
patrickhart: I think we can discuss how we could unit test this next week.
Then it should be possible to still implement it until Thursday!

patrickhart:
> Please consider writing a unit test for ChooseLaneCorridorBasedOnVehicleState. Also, did you check if BehaviorLaneChangeMobil is still working (in the examples)?

For now the selection is only active if a lane corridor has been set. So should not affect the behavior of BARK overall.
I think a system test validation is for now sufficient.

bark/world/map/commons.hpp
patrickhart: No

patrickhart:
Integrated in branch `rss_benchmark`.

PR #483, opened 2020-12-04
Rss develop

tobiaskessler:


bark/models/behavior/idm/base_idm.cpp
```diff
@@ -123,7 +124,7 @@ double BaseIDM::CalcNetDistance(
 
   // we need to use the lane corridor of the ego agent to be able to compare the
   // frenet values
-  const auto& lane_corridor = observed_world.GetLaneCorridor();
+  const auto& lane_corridor = local_lane_corr;
```
patrickhart, line 127: Why not just use local_lane_corr in the following?

patrickhart (approved):

bark/models/behavior/idm/base_idm.cpp
tobiaskessler: true.

PR #497, opened 2021-03-19
Parametric observer

juloberno:
- parametric observer implementation

### DIFF
--- a/bark/models/observer/observer_model_parametric.cpp
+++ b/bark/models/observer/observer_model_parametric.cpp
@@ -1,0 +1,81 @@
+// Copyright (c) 2020 fortiss GmbH
+//
+// Authors: Julian Bernhard, Klemens Esterle, Patrick Hart and
+// Tobias Kessler
+//
+// This work is licensed under the terms of the MIT license.
+// For a copy, see <https://opensource.org/licenses/MIT>.
+
+
+#include <memory>
+#include "bark/models/observer/observer_model_parametric.hpp"
+#include "bark/commons/distribution/distribution.hpp"
+#include "bark/commons/transformation/frenet_state.hpp"
+#include "bark/world/observed_world.hpp"
+
+namespace bark {
+namespace models {
+namespace observer {
+using namespace bark::world;
+using namespace bark::world::objects;
+using namespace bark::commons::transformation;
+using namespace bark::commons;
+
+ObserverModelParametric::ObserverModelParametric(bark::commons::ParamsPtr params) :
+  ObserverModel(params),
+  ego_state_deviation_dist_(params->GetDistribution("ObserverModelParametric::EgoStateDeviationDist",
+                                  "From what distribution is the ego frenet state deviations "
+                                  "sampled, must have dimension = 5",
+                                  "MultivariateDistribution")),
+  others_state_deviation_dist_(params->GetDistribution("ObserverModelParametric::OtherStateDeviationDist",
+                                  "From what distribution is the others frenet state deviation"
+                                  "sampled, must have dimension = 5",
+                                  "MultivariateDistribution")) {}
+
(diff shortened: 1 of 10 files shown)


bark/models/observer/observer_model_parametric.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+
+ObservedWorld ObserverModelParametric::Observe(
+    const WorldPtr& world, const AgentId& agent_id) {
+  
+  // Clone world here since otherwise we change global world state
+  auto observed_world = ObservedWorld(world->Clone(), agent_id);
+
+  AddStateDeviationFrenet(observed_world.GetEgoAgent(), ego_state_deviation_dist_);
+  for (auto& agent : observed_world.GetOtherAgents()) {
```
patrickhart, line 48: do we only want the possibility of the other agents to be perturbed or should we also introduce it for the ego-agent, e.g. modeling a poor localization?

bark/models/observer/observer_model_parametric.cpp
```diff
@@ -0,0 +1,81 @@
...
+  current_frenet_state.lat += frenet_deviation[1];
+  current_frenet_state.vlat += frenet_deviation[2];
+  current_frenet_state.vlon += frenet_deviation[3];
+  current_frenet_state.angle += frenet_deviation[4];
+  
+  // Convert back and set dynamic agent state 
+  const auto deviated_state = FrenetStateToDynamicState(current_frenet_state, lane_corridor->GetCenterLine());
+  agent->SetCurrentState(deviated_state);
+
+  return;
```
patrickhart, line 76: is a return in a void really required? better to return a boolean instead maybe?

bark/models/observer/observer_model_parametric.hpp
```diff
@@ -0,0 +1,65 @@
...
+using bark::commons::DistributionPtr;
+
+/**
+ * @brief  Observer creating an ObservedWorld given a World
+ * @note   
+ * @retval None
+ */
+class ObserverModelParametric : public ObserverModel {
+ public:
+  explicit ObserverModelParametric(bark::commons::ParamsPtr params);
```
patrickhart, line 45: add `using bark::commons::ParamsPtr` above and pass as const ref: `const ParamsPtr& params`

bark/world/objects/agent.hpp
```diff
@@ -170,6 +170,10 @@ class Agent : public Object {
     history_ = history;
   }
 
+  void SetCurrentState(const State& current_state) {
+    history_.back().first = current_state;
```
patrickhart, line 174: should we not fill the action also with an default value?

patrickhart (approved):

bark/models/observer/observer_model_parametric.cpp
juloberno: See line 47, where this is implemented

bark/models/observer/observer_model_parametric.cpp
juloberno: I check the conventions

bark/world/objects/agent.hpp
juloberno: Yeah, thats right there is missing the case of the history being empty. But the function is only called after the constructor where the first entry of the history is always set

bark/models/observer/observer_model_parametric.hpp
juloberno: okay

bark/world/objects/agent.hpp
juloberno: Yeah, thats right there is missing the case of the history being empty. But the function is only called after the constructor where the first entry of the history is always set

PR #500, opened 2021-03-25
added bernoulli distribution

amitsahu7:
Part of Implementation of Discrete Probability Distributions


patrickhart:
CI needs to be triggered. See PR #501

PR #504, opened 2021-03-28
Renderer in World

patrickhart:
The Renderer class enables adding (geometric) primitives to be drawn.

---

It implements a `buffer_` that, e.g., stores  geometric primitives, such as Lines, Polygons, etc.
This buffer can be filled either by adding primitives manually or by using the BufferedViewer.

The Renderer class itself does not implement any rendering functionality (yet).
It can be rendered using any kind of framework, such as, e.g., Unity.

---

Also removed the Pandas3d viewer due to inactivity.


bark/examples/BUILD
```diff
@@ -124,7 +120,6 @@ py_binary(
           "//bark/runtime/commons:xodr_parser",
           "//bark/runtime/viewer:pygame_viewer",
           "//bark/runtime/viewer:matplotlib_viewer",
-          "//bark/runtime/viewer:panda3d_easy",
```
klemense1, line 127: ich würde sagen wir schmeißen den panda viewer ganz weg. den nutzen wir eh nie...

bark/runtime/viewer/buffered_viewer.py
```diff
@@ -0,0 +1,88 @@
...
+import numpy as np
+from bark.core.geometry import *
+from bark.core.models.dynamic import *
+from bark.core.world.opendrive import *
+from bark.core.world.goal_definition import *
+from bark.core.world.renderer import *
+from bark.runtime.commons.parameters import ParameterServer
+
+
+class BufferedViewer:
```
klemense1, line 18: what's the difference to the Viewer Class? will we maintain both? or should the viewers inherit from this?

klemense1 (approved):
Looks good!

bark/examples/BUILD
patrickhart: hab ich gemacht!

bark/runtime/viewer/buffered_viewer.py
patrickhart: This BufferedViewer just writes (geometric) primitives into the renderer's buffer. The buffer can then, e.g., be visualized using BARKSCAPE.

bark/world/world.hpp
```diff
@@ -194,6 +200,7 @@ class World : public commons::BaseType {
   MapInterfacePtr map_;
   AgentMap agents_;
   ObjectMap objects_;
+  RendererPtr renderer_;
```
juloberno, line 203: Why does the renderer have to be a member of the world class?

juloberno (changes requested):
Looks good, apart from the implementation detail requiring another member of the world class

juloberno (approved):
Looks good, after clarifying that a pointer-based implementation serves as basis for implemented a viewer hierarchy with child viewers

patrickhart:
Okay, I will merge this then for now. And add the child renderer functionality in a separate PR.

PR #507, opened 2021-03-30
Integration of the ObserverModel into the ConfigurableScenario generation

patrickhart:
Include the `ObserverModel` into the `ConfigurableScenarioGeneration`

---

- Added new config reader `ConfigReaderObserverModel`
- Added `ObserverModelNoneReader` and `ObserverModelNoneReaderParametric`
- integration of new config readers into `ConfigurableScenarioGeneration`


bark/runtime/scenario/scenario_generation/config_readers/observer_model_config_readers.py
```diff
@@ -0,0 +1,38 @@
...
+    params = ParameterServer()
+    observer_model = ObserverModelNone(params)
+    return observer_model, {}, config_param_object
+
+# observed worlds with parametric perturbations
+class ObserverModelParametricReader(ConfigReaderObserverModel):
+  # returns a parametric observer model for the world
+  def create_from_config(self, config_param_object, **kwargs):
+    params = ParameterServer()
+    # if "EgoStateDeviationDist" in config_param_object:
```
patrickhart, line 31: @juloberno is there a convenient way to generate a distribution here? what in your point of view would be the best implementation solution?

patrickhart (commented):
Pre-final implementation.

bark/runtime/scenario/scenario_generation/config_readers/observer_model_config_readers.py
```diff
@@ -0,0 +1,38 @@
...
+from bark.runtime.scenario.scenario_generation.config_readers.config_readers_interfaces import ConfigReaderObserverModel
+from bark.runtime.commons.parameters import ParameterServer
+from bark.core.models.observer import ObserverModelNone, ObserverModelParametric
+
+
+# observed worlds with no perturbations
+class ObserverModelNoneReader(ConfigReaderObserverModel):
+  # returns a observer model none for the world
+  def create_from_config(self, config_param_object,  **kwargs):
+    params = ParameterServer()
```
juloberno, line 22: You should not create an own param server here, but just use the config param object. Have a look at the other readers how to pass down the param server

bark/runtime/scenario/scenario_generation/config_readers/observer_model_config_readers.py
juloberno: You can just create the ParametricObserver and pass the config_param_object, distributions are all internally managed by the observer

juloberno (changes requested):
In principle, looks good. Only the readers need small adjustments.

bark/runtime/scenario/scenario_generation/config_readers/observer_model_config_readers.py
patrickhart: yeah, that was a workaround

PR #510, opened 2021-04-01
Observed world viewer

juloberno:


bark/models/observer/observer_model.hpp
```diff
@@ -39,10 +39,12 @@ using world::WorldPtr;
 class ObserverModel : public commons::BaseType {
  public:
   explicit ObserverModel(bark::commons::ParamsPtr params)
-    : BaseType(params) {}
+    : BaseType(params),
+      observe_only_for_agents_() {}
```
patrickhart, line 43: would rather name it `observed_agent_ids_`. Avoids confusion with using pointers

bark/models/observer/observer_model.hpp
```diff
@@ -55,6 +57,17 @@ class ObserverModel : public commons::BaseType {
   virtual ObservedWorld Observe(
     const WorldPtr& world, const AgentId& agent_id) = 0;
 
+  void SetObserveOnlyForAgents(const std::vector<AgentId>& observe_only_for_agents) {
```
patrickhart, line 60: `SetObservedAgentIds` respectively

bark/models/observer/observer_model_parametric.cpp
```diff
@@ -44,6 +44,12 @@ ObservedWorld ObserverModelParametric::Observe(
   // Clone world here since otherwise we change global world state
   auto observed_world = ObservedWorld(world->Clone(), agent_id);
 
+  const auto observe_only_for = GetObserveOnlyForAgents();
+  if(!observe_only_for.empty() && std::find(observe_only_for.begin(),
```
patrickhart, line 48: `obsrved_agent_ids` respectively

bark/python_wrapper/models/observer.cpp
```diff
@@ -40,7 +42,7 @@ void python_observer(py::module m) {
 
   py::class_<ObserverModelParametric, ObserverModel,
              shared_ptr<ObserverModelParametric>>(m, "ObserverModelParametric")
-    .def(py::init<const bark::commons::ParamsPtr&>())
+    .def(py::init<const bark::commons::ParamsPtr&> ())
```
patrickhart, line 45: why space?

bark/runtime/runtime.py
```diff
@@ -53,10 +53,11 @@ def step(self):
     if len(self._world_history) == 0 and self._maintain_world_history:
         self._world_history.append(self._world.Copy())
 
-    self._world.Step(self._step_time)
+    self._world.PlanAgents(self._step_time)
 
     if self._render:
```
patrickhart, line 58: render before the execution? I do not think that is a good idea?

bark/runtime/runtime.py
```diff
@@ -53,10 +53,11 @@ def step(self):
...
         self.render()
+    self._world.Execute(self._step_time)
```
patrickhart, line 60: That means that render always plots the old world state.

bark/world/world.cpp
```diff
@@ -70,6 +70,7 @@ void World::PlanAgents(const double& delta_time) {
     if (agent.second->IsValidAtTime(world_time_)) {
       ObservedWorld observed_world = observer_->Observe(
         current_world, agent.first);
+      agent.second->SetSensedWorld(std::make_shared<ObservedWorld>(observed_world));
```
patrickhart, line 73: Isn't a `SensedWorld` also an `ObservedWorld`?

patrickhart: This might cause confusion

bark/world/objects/agent.hpp
```diff
@@ -193,6 +202,8 @@ class Agent : public Object {
   uint32_t max_history_length_;
   GoalDefinitionPtr goal_definition_;
   double first_valid_timestamp_;
+
+  ObservedWorldPtr sensed_world_;
```
patrickhart, line 206: Maybe `sampled_observed_world_` or something like this would be better.
How about the concept that we wanted to have two observed worlds? One on the simulation side and one on the agent's side?

At the moment the one from the simulation is set as the one of the agent as well?

bark/runtime/viewer/viewer.py
```diff
@@ -289,6 +289,47 @@ def drawLabelsAsText(self, observed_world, label_functions, evaluator_type):
...
+          self.drawWorldImplementation(world, eval_agent_ids, filename, scenario_idx, debug_text)
+
+    def drawSensedWorld(self, sensed_world, eval_agent_ids=None, filename=None, scenario_idx=None, debug_text=True):
+      self.drawWorldImplementation(sensed_world, eval_agent_ids, filename, scenario_idx, debug_text)
+
+    def drawTrueWorld(self, world, eval_agent_ids):
+      # draw only boundaries of agents in same color as observed agents
+      for i, (agent_id, agent) in enumerate(world.agents.items()):
+          color_line, color_face, alpha = self.GetAgentColor(agent, eval_agent_ids)
+          color_face = (1.0, 1.0, 1.0) # face color white
```
patrickhart, line 306: why set a default color here?

patrickhart (commented):
Have a look over the comments

patrickhart:
Also, even though we know what the PR is for, I think a description would be beneficial so others can trace the changes as well

bark/runtime/runtime.py
juloberno: Yes, thats a good idea:
- It allows to draw planned mcts states or other information from the planner, ontop of the world state it was planned for
- It allows to draw the correct senses world ontop of the true world

bark/world/world.cpp
juloberno: Later on we may introduce the perceived world which is what a perception component will provide to the agent. The sensed world is then passed to the perception component to generate a perceived world. Therefore, we need to differentiate though it is basically the same type. I do not know if it makes sense to define separate derived types.

bark/world/objects/agent.hpp
juloberno: see comment above

bark/runtime/viewer/viewer.py
juloberno: we ground truth shape should only have borders drawn and color white. I did not want to introduce another parameter when it does not make sense for me.

bark/runtime/runtime.py
patrickhart: As first step in BARK we call runtime.reset() which returns `s_0`.

Then we render the world after `PlanAgents` which still returns `s_0`.

So de facto, `s0` is then rendered twice. Is this intended?

bark/world/world.cpp
patrickhart: I think the difference between perceived and sensed world is difficult to grasp.

`SimulatedWorld` and `SensedWorld` would enlarge the gap.

So the `SimulatedWorld` world could model some uncertainties according to a specified (unknown to the agent) distribution and the agent has a `SensedWorld` that it, e.g., could use in a Kalman filter.

bark/runtime/runtime.py
klemense1: I agree with Julian

bark/world/objects/agent.hpp
klemense1: why do we need to store it as a member? what is missing with the current implementation of passing it as an argument of the Plan function?

bark/examples/parametric_observer.py
```diff
@@ -0,0 +1,136 @@
...
+  env.reset()
+  # step each scenario 20 times
+  for step in range(0, 40):
+    env.step()
+    time.sleep(sim_step_time/sim_real_time_factor)
+    
+  df = env.ExtractTimeSeries()
+  print(df)
+
+viewer.export_video(filename="/home/julo/video", remove_image_dir=False)
```
klemense1, line 136: use some relative path name

klemense1 (approved):

patrickhart (approved):

bark/models/observer/observer_model.hpp
juloberno: Mhh, I think observe_only_for_agents_ is clearer since it describes what to is done with this information. observed_agent_ids_

bark/models/observer/observer_model_parametric.cpp
juloberno: see previous comment

bark/runtime/runtime.py
juloberno: We have to find a way to draw the last state on reset()

bark/world/world.cpp
juloberno: The questions is do we need an extra type definition or do we just name them differently. I do not understand what you prefer yet.

juloberno: I think we can still rename things and stuff, but there is no way around an agent holding a pointer from its last observed world. If we want to persist the information somewhere its the best place for me.

bark/world/objects/agent.hpp
juloberno: We cannot plot it in the viewer since the observed_world is nowhere persisted.

PR #519, opened 2021-04-12
Class Constructor for Line Geometry and Time Measurement of Behavior Plan Function

klemense1:
* Timer Class using Boost
* using already existing solution time member and it's respective evaluator
* additional constructor for line class to prevent repetitively calling calculation of s-value with every new point

### DIFF
--- a/bark/commons/timer/timer.hpp
+++ b/bark/commons/timer/timer.hpp
@@ -1,0 +1,37 @@
+// Copyright (c) 2021 fortiss GmbH, Julian Bernhard, Klemens Esterle, Patrick
+// Hart, Tobias Kessler
+//
+// This work is licensed under the terms of the MIT license.
+// For a copy, see <https://opensource.org/licenses/MIT>.
+
+#ifndef BARK_COMMONS_TIMER_TIMER_HPP_
+#define BARK_COMMONS_TIMER_TIMER_HPP_
+
+#include "boost/date_time/posix_time/posix_time.hpp"
+
+namespace bark {
+namespace commons {
+namespace timer {
+
+class Timer {
+ public:
+  Timer() : t1_() {}
+  virtual ~Timer(){};
+  inline void Start() { t1_ = boost::posix_time::microsec_clock::local_time(); }
+
+  inline double DurationInSeconds() {
+    boost::posix_time::ptime t2 =
+        boost::posix_time::microsec_clock::local_time();
+    boost::posix_time::time_duration diff = t2 - t1_;
+    return static_cast<double>(diff.total_milliseconds()) / 1e3;
+  }
+
+ private:
+  boost::posix_time::ptime t1_;
+};
+
+}  // namespace timer
+}  // namespace commons
+}  // namespace bark
+
+#endif  // BARK_COMMONS_TIMER_TIMER_HPP_
--- a/bark/commons/timer/timer.hpp
+++ b/bark/commons/timer/timer.hpp
@@ -23,7 +23,7 @@ class Timer {
     boost::posix_time::ptime t2 =
         boost::posix_time::microsec_clock::local_time();
     boost::posix_time::time_duration diff = t2 - t1_;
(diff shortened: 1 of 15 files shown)


bark/commons/timer/timer.hpp
```diff
@@ -0,0 +1,37 @@
...
+ public:
+  Timer() : t1_() {}
+  virtual ~Timer(){};
+  inline void Start() { t1_ = boost::posix_time::microsec_clock::local_time(); }
+
+  inline double DurationInSeconds() {
+    boost::posix_time::ptime t2 =
+        boost::posix_time::microsec_clock::local_time();
+    boost::posix_time::time_duration diff = t2 - t1_;
+    return static_cast<double>(diff.total_milliseconds()) / 1e3;
```
patrickhart, line 26: why divide it my 1e3? Can use, e.g., `total_seconds`

bark/world/objects/agent.cpp
```diff
@@ -73,7 +74,11 @@ Agent::Agent(const Agent& other_agent)
 
 void Agent::PlanBehavior(const double& min_planning_dt,
                          const ObservedWorld& observed_world) {
+  auto timer = bark::commons::timer::Timer();
+  timer.Start();
   behavior_model_->Plan(min_planning_dt, observed_world);
+  double duration = timer.DurationInSeconds();
+  behavior_model_->SetLastSolutionTime(duration);
```
patrickhart, line 81: I would rather set the last time the behavior has been planned, and e.g. return a logging message if the planning time was larger then the step time dt (e.g., 0.2s)

patrickhart: This would also enable replanning after a specific time

patrickhart (commented):
Looks good, see comments

bark/world/objects/agent.cpp
patrickhart: Or the behavior model itself could handle the timer as, e.g., the IDM does not need to be timed.

bark/commons/timer/timer.hpp
klemense1: no, then the resolution is only in seconds

bark/world/objects/agent.cpp
klemense1: the time measurement is rather for the benchmarking, I would say. In the end, we will be able to evaluate pretty easily of the planning is real-time capable or not. Thus, I would prefer to save the time. We could move it to the behavior model (that's what I did for MIQP already, but @juloberno wish was to make this available for every model in a generic way. I don't think that it adds much overhead, but we could introduce a flag in the behavior model, that states wether time should be measured or not. In the end, this flag could be set only for the evaluation agent, what do you think?

juloberno: I think it is a meaningful feature, but the implementation is confusing. 
If the agent measures timing, then the agent should have the planning time. Otherwise the behavior model should handle this internally with a redefinition of the Plan interface and and an additional purely virtual PlanImplementation function.

Also, it should be possible to avoid overhead, e.g. using a parameter. Maybe, then still the behavior-model specific solution is better.

juloberno (changes requested):

bark/world/objects/agent.cpp
patrickhart: what is the status here, should we agree on a solution?

I think it would be interesting to see how long the model runs in general and how efficient it is. Just wanted to note that if we saved a timestamp when the behavior/execution model last planned a trajectory, we could enable replanning after a specific time.

klemense1: I've already implemented something, I will add a test and then push it.

juloberno (approved):
Finde ich eine gute Lösung: flexibel je Planer und dennoch allgemein parametrierbar

patrickhart (approved):
Passt von meiner Seite auch

PR #523, opened 2021-05-04
Observer parametric velocity deviation

juloberno:
- switch back to non frenet-based state deviation
- makes actually more sense since sensor noise is not influence by the current lane the vehicle is but instead mainly by the direction the object is moving to

### DIFF
--- a/bark/models/observer/observer_model_parametric.cpp
+++ b/bark/models/observer/observer_model_parametric.cpp
@@ -55,10 +55,10 @@ ObservedWorld ObserverModelParametric::Observe(
       delta_time = world->GetWorldTime() - previous_observed_world->GetWorldTime();
   }
   AddStateDeviationFrenet(observed_world.GetEgoAgent(), ego_state_deviation_dist_,
-                        previous_observed_world, delta_time);
+                        previous_observed_world);
   for (auto& agent : observed_world.GetOtherAgents()) {
     AddStateDeviationFrenet(agent.second, others_state_deviation_dist_,
-                            previous_observed_world, delta_time);
+                            previous_observed_world);
   }
 
   return observed_world;
--- a/bark/models/observer/observer_model_parametric.cpp
+++ b/bark/models/observer/observer_model_parametric.cpp
@@ -65,7 +65,7 @@ ObservedWorld ObserverModelParametric::Observe(
 }
 
 void ObserverModelParametric::AddStateDeviationFrenet(const AgentPtr& agent, const DistributionPtr& multi_dim_distribution,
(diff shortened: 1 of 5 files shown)


bark/models/observer/observer_model_parametric.cpp
```diff
@@ -49,36 +49,36 @@ ObservedWorld ObserverModelParametric::Observe(
...
-  // Get Current Frenet State of Agent
-  const Point2d pos = agent->GetCurrentPosition();
-  const auto& lane_corridor = agent->GetRoadCorridor()->GetCurrentLaneCorridor(pos);
-  if(!lane_corridor) {
-    return;
-  };
-  FrenetState current_frenet_state(agent->GetCurrentState(), lane_corridor->GetCenterLine());
-
+void ObserverModelParametric::AddStateDeviationFrenet(const AgentPtr& agent, const DistributionPtr& multi_dim_distribution,
+                         const ObservedWorldPtr& previous_observed_world, const double& delta_time) const {
```
patrickhart, line 68: If we do not use delta_time, do we need this here?

bark/models/observer/observer_model_parametric.cpp
```diff
@@ -49,36 +49,36 @@ ObservedWorld ObserverModelParametric::Observe(
...
-  current_frenet_state.lon += frenet_deviation[0];
-  current_frenet_state.lat += frenet_deviation[1];
-  
-  // Convert back and set dynamic agent state 
-  const auto deviated_state = FrenetStateToDynamicState(current_frenet_state, lane_corridor->GetCenterLine());
-  agent->SetCurrentState(deviated_state);
+  // Get Current Frenet State of Agent
+  auto state = agent->GetCurrentState();
+  const auto theta = state[StateDefinition::THETA_POSITION];
+  state[StateDefinition::X_POSITION] += state_deviation[0]*cos(theta) + state_deviation[1]*sin(theta);
```
patrickhart, line 76: why do we need a transformation here? we can have a distribution that defines the deviations on the Cartesian coordinates

patrickhart: Unless theta is used from the sampled values. so would need to update theta before and then calculate the x, y position

juloberno: Deviations in Cartesian coordiantes would change the results if we rotate the whole scene. This formulation is independent of road placement in world coordinates.

patrickhart: but theta would still be sampled, why not use the sampled one?

patrickhart (approved):

PR #530, opened 2021-05-19
Using pybind_bazel and build wheels for Linux/macOS(Python 3.7, 3.8, 3.9)

patrickhart:
Requiring a local virtual environment folder restricts platform independence and hinders building BARK for multiple Python versions. The majority of Python users now use Python 3.8 (https://www.jetbrains.com/lp/python-developers-survey-2020/) which BARK should be compatible for.

In this pull request:

* Using `pybind_bazel ` avoids defining a local virtual python environment for building the python wrappings.
* Extended `setup.py` using [Bazelisk](https://github.com/bazelbuild/bazelisk) making building BARK easy even without having installed Bazel 
* Integrated GitHub Actions wheel builds that build BARK for Python3.7, 3.8, and 3.9 on Linux and macOS making it possible to easily use BARK using `pip install bark-simulator`
* wheels are only built and pushed when a git tag is pushed on the CI
* cleanup of the build structure; for some reason BARK required BARK itself as dependency; etc.
* removed manylinux CI build
* added MANIFEST.in to only include required files (reduces wheel size)
* generated wheels can be found here: https://pypi.org/project/bark-simulator/#files

Current problems not easily fixable:
* macOS runners reach rate limit on http request (error 403) on Bazelisk call. seems to be a more common problem. did not find a fix. results in arbitrary passing/failing of the macOS wheel builds


bark/benchmark/BUILD
```diff
@@ -1,7 +1,7 @@
 py_library(
     name = "benchmark_runner",
     srcs = ["benchmark_runner.py"],
-    data = ['//bark:generate_core'],
+    data = ["//bark:generate_core"],
     imports = ['../../../python'],
```
klemense1, line 5: make those quotation marks congurent? always use "?

tools/deps.bzl
```diff
@@ -3,10 +3,10 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository", "new_git_r
...
 def bark_dependencies():
     _maybe(
-    git_repository,
-    name = "bark_project",
-    branch = "master",
-    remote = "https://github.com/bark-simulator/bark"
+        git_repository,
+        name = "pybind11_bazel",
+        commit="378ce0523b85ad1748cad49ad9b739a91000e81c",
+        remote = "https://github.com/patrickhart/pybind11_bazel"
```
klemense1, line 9: why do you use your fork?

bark/world/tests/evaluation/BUILD
```diff
@@ -17,7 +17,7 @@ py_test(
         "//conditions:default": ["py_dummy_test.py"],
     }),
     data = [
-        "//bark/python_wrapper:core.so",
```
klemense1, line 20: how is it possible to remove /python_wrapper/ in the path?

klemense1 (approved):

bark/benchmark/BUILD
patrickhart: yes, that's why I changed it!

tools/deps.bzl
patrickhart: the main repo ist not being maintained and no pull requests are accepted. There is an issue that describes how to "fix" the error

klemense1: then let's fork it as bark_simulator?

PR #546, opened 2021-09-30
Junction area as part of the road corridor

tobiaskessler:
On junctions, the road corridor was rather a lane corridor. This in intersection scenarios somehow predefined the interaction of vehicles as the set of actions in the intersection was very limited.

I now added ALL lanes inside a junction to a junction area that is used for the road corridor, if it passes the junction, see screenshot.

Default behavior is UNCHANGED, shouldnt break anything. Activate in the map interface via:
`map_interface = MapInterface()
map_interface.full_junction_area = True
map_interface.SetOpenDriveMap(xodr_parser.map)`
see also py_map_interface_test

Screenshot from 2021-09-30 09-12-11

### DIFF
--- a/bark/python_wrapper/world/map.cpp
+++ b/bark/python_wrapper/world/map.cpp
@@ -26,6 +26,9 @@ using bark::world::opendrive::XodrLaneId;
 void python_map(py::module m) {
   py::class_<MapInterface, std::shared_ptr<MapInterface>>(m, "MapInterface")
       .def(py::init<>())
+      .def_property("add_full_junction_area",
+                    &MapInterface::GetAddFullJunctionArea,
+                    &MapInterface::SetAddFullJunctionArea)
       .def("SetOpenDriveMap", &MapInterface::SetOpenDriveMap)
       .def("find_nearest_lanes",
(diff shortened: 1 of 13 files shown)


bark/python_wrapper/world/map.cpp
```diff
@@ -26,6 +26,9 @@ using bark::world::opendrive::XodrLaneId;
 void python_map(py::module m) {
   py::class_<MapInterface, std::shared_ptr<MapInterface>>(m, "MapInterface")
       .def(py::init<>())
+      .def_property("add_full_junction_area",
```
klemense1, line 29: I'm a little bit confused about the naming. is it really a property? then remove "add"?

klemense1 (commented):
seems fine for me besides the naming of the python wrapping functionality

bark/python_wrapper/world/map.cpp
tobiaskessler: its a boolean flag if the junction area shall be added to the road corridor. I renamed it.

patrickhart (approved):
Looks good to me! And, true, before is was just a LaneCorridor.

PR #562, opened 2021-12-04
Single Track model with Steering-Rate Inputs

patrickhart:
Added a SingleTrackModel that takes the acceleration and steering-rate as input.


bark/models/dynamic/single_track_steering_rate.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+  State StateSpaceModel(const State& x, const Input& u) const {
+    State tmp(static_cast<int>(StateDefinition::MIN_STATE_SIZE ) + 1);
+    tmp << 1,
+        x(StateDefinition::VEL_POSITION) *
+            cos(x(StateDefinition::THETA_POSITION)),
+        x(StateDefinition::VEL_POSITION) *
+            sin(x(StateDefinition::THETA_POSITION)),
+        x(StateDefinition::VEL_POSITION) * tan(
+          x[5]) / wheel_base_,
```
klemense1, line 40: x[5] should also use StateDefinition

bark/models/dynamic/single_track_steering_rate.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include "bark/commons/transformation/frenet_state.hpp"
+#include "bark/models/dynamic/dynamic_model.hpp"
+
+namespace bark {
+namespace models {
+namespace dynamic {
+
+
+class SingleTrackSteeringRateModel : public DynamicModel {
+ public:
```
klemense1, line 23: please add some doxygen documentation here... it is not obvious why the model consists of...

bark/models/tests/behavior_dynamic_model.cc
```diff
@@ -0,0 +1,106 @@
+// Copyright (c) 2020 fortiss GmbH
```
klemense1, line 1: rename the file test_...?

klemense1 (changes requested):
Looks good, please follow up on my remarks.

bark/models/dynamic/single_track_steering_rate.hpp
patrickhart: The problem is that at the moment `StateDefinition::MIN_STATE_SIZE` is equal to `5`. So I don't think we can change that with major refactoring.

klemense1: Okay, although I don't like it :-)

klemense1 (approved):

